PR #1, opened 2019-02-09
Migrate the ViewPager module from React Native core to the community repo

ferrannp:
## Done

- [x] Keep git history from RN core
- [x] Changed module name to `RNCViewPager`
- [x] Fix bug on the example: [This from RNTester](https://github.com/facebook/react-native/blob/master/RNTester/js/ViewPagerAndroidExample.android.js#L115-L117) should be like [here](https://github.com/ferrannp/react-native-viewpager/blob/init-with-history/example/ViewPagerAndroidExample.android.js#L113-L115)

## ToDo after merging this

- Add eslint + prettier
- Add CI
- Deprecate and remove from RN core repo
- Refactor JS code a bit?
- Fix warnings and deprecations from Android
- Modernize Android code (even migrate to Kotlin if we want to)

## Example

viewpager


example/ViewPagerAndroidExample.android.js
```diff
@@ -112,8 +110,8 @@ class ViewPagerAndroidExample extends React.Component {
     this.setState({progress: e.nativeEvent});
   };
 
-  onPageScrollStateChanged = (state: ViewPagerScrollState) => {
-    this.setState({scrollState: state});
+  onPageScrollStateChanged = e => {;
```
cpojer: you got a semicolon here that shouldn't be here :D

js/index.js
```diff
@@ -0,0 +1,3 @@
+/* @flow */
+
+export { default } from './ViewPagerAndroid';
```
cpojer: pls remove the spaces around the curly braces.

PR #2, opened 2019-02-11
measure and layout after adding a view to the container

danilobuerger:
Imported from https://github.com/facebook/react-native/pull/18609
---

Currently, when adding views to ViewPagerAndroid after it was attached, those views would not show up as they had a size of 0. This fixes that by measuring and layouting after adding a view to the container.

Fixes https://github.com/facebook/react-native/issues/14296

## Test Plan

See https://github.com/danilobuerger/react-native-viewpagerandroid-dynamic

1. yarn install
2. yarn run android
3. You will see two pages
4. Press "ADDPAGE"
5. Scroll to the third page, it will be empty
6. Use react-native master with this patch
7. yarn run android
8. You will see two pages
9. Press "ADDPAGE"
10. Scroll to the third page, it will show "3 page (Scroll Me)"

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/viewpager/ReactViewPager.java
+++ b/android/src/main/java/com/reactnativecommunity/viewpager/ReactViewPager.java
@@ -36,24 +36,11 @@ public class ReactViewPager extends ViewPager {
     void addView(View child, int index) {
       mViews.add(index, child);
       notifyDataSetChanged();
-      // This will prevent view pager from detaching views for pages that are not currently selected
-      // We need to do that since {@link ViewPager} relies on layout passes to position those views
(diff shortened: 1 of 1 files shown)


satya164:
Thanks ❤️

ferrannp:
Awesome! Will merge this after making the first release. After that, we can add your Add Page button to the example here too :)

danilobuerger:
@ferrannp Thanks! Could you also cut a new release with this fix please? I guess this repo won't have much activity for the first few weeks until people start migrating, so it would be nice if its already there.

ferrannp:
Yup, thank you!

danilobuerger:
@ferrannp Awesome, thanks!

PR #9, opened 2019-02-14
compat with <= rn57

danilobuerger:
`UIManager.getViewManagerConfig` does not exist on RN <= 0.57. Added fallback for that case.

### DIFF
--- a/js/ViewPagerAndroid.android.js
+++ b/js/ViewPagerAndroid.android.js
@@ -23,6 +23,14 @@ const NativeAndroidViewPager = require('./AndroidViewPagerNativeComponent');
 
 const VIEWPAGER_REF = 'viewPager';
 
+function getViewManagerConfig(viewManagerName) {
+  if (!UIManager.getViewManagerConfig) {
+    // react-native <= 0.57
+    return UIManager[viewManagerName];
+  }
+  return UIManager.getViewManagerConfig(viewManagerName);
+}
+
 type PageScrollState = 'idle' | 'dragging' | 'settling';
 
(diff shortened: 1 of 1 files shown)


ferrannp:
Hey @danilobuerger, do you know what peer dependency should we use then? Wondering for how long this have been working :D gonna investigate on that.

danilobuerger:
I am not sure, but this specific detail should have been available for a really long time. At least 3 years. See https://github.com/facebook/react-native/commit/60db876f666e256eba8527251ce7035cfbca6014

js/ViewPagerAndroid.android.js
```diff
@@ -12,7 +12,7 @@
 
 const React = require('react');
 const ReactNative = require('react-native');
-const {UIManager} = ReactNative;
+const {UIManager} = ReactNative.NativeModules;
```
ferrannp: Why would you change the way we import it? I see in the docs how they:

```
import { UIManager } from 'react-native';
```

danilobuerger: Docs might not be in sync with actual code base. See for example:
https://github.com/facebook/react-native/blob/master/Libraries/ReactNative/UIManager.js#L19

danilobuerger: Yes you are right, should be
```
import { UIManager } from 'react-native';
```
will amend.

danilobuerger: @ferrannp amended.

danilobuerger:
Thanks @ferrannp ! Would it be possible to cut a new release please?

PR #14, opened 2019-03-13
Add .npmignore and fix compatibility with Jest

### LINKED ISSUE #12: Package size is huge
It seems like the example dir is bundled within the npm package resulting in a huge package size.

### LINKED ISSUE #13: Fails with jest
Using `ViewPagerAndroid` within a test fails with:

```
  ● Test suite failed to run

    Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    node_modules/@react-native-community/viewpager/js/index.js:3
    export {default} from './ViewPagerAndroid';
    ^^^^^^

    SyntaxError: Unexpected token export



      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/script_transformer.js:403:17)
      at Object.<anonymous>
```

Adding the repos to `transformIgnorePatterns` solves the problem. But wouldn't it be better to transpile this to ES5 when bundling?

ferrannp:
* Fixes #12 (no need to add what we already have in `.gitignore`
* Fixes #13

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,3 @@
+/.circleci
+/example
+babel.config.js
--- a/js/index.js
+++ b/js/index.js
@@ -1,3 +1,3 @@
 /* @flow */
 
-export {default} from './ViewPagerAndroid';
+module.exports = require('./ViewPagerAndroid');


danilobuerger:
@ferrannp @cpojer Would it be possible to cut a new release with this please? Then we don't have a massive dependency to download.

ferrannp:
You got it.

PR #17, opened 2019-03-16
Exclude viewpager.gif to reduce bundle size

danilobuerger:
Ref #12


danilobuerger:
@ferrannp The `viewpager.gif` was still being bundled.

Issue #20, opened 2019-04-12
CocoaPod support


congnguyen91:
- Hi, bring pod spec file to git, home, please :) 
Thanks

uinz:
iOS supported?

PR #21, opened 2019-04-15
Fixes podspec errors.

kkoudev:
Fix: #20 

Fixes podspec errors.
Some attributes are referenced from package.json.


ferrannp:
Thank you! ❤️

PR #22, opened 2019-04-27
Update Android project to use AndroidX support library

aarongrider:
Use AndroidX in place of legacy Android support libraries. This is required for any Android projects that use AndroidX for support libraries.

See: https://developer.android.com/jetpack/androidx/migrate

Specifically react-native-gesture-handler currently has a 4 PRs to migrate to AndroidX being blocked by this down stream dependency for our tests that require this dependency.

### DIFF
--- a/example/android/build.gradle
+++ b/example/android/build.gradle
@@ -5,7 +5,7 @@ buildscript {
         minSdkVersion = 16
         compileSdkVersion = 28
         targetSdkVersion = 27
-        supportLibVersion = "28.0.0"
+        supportLibVersion = "1.0.2"
     }
     repositories {
         google()
--- a/android/gradle.properties
+++ b/android/gradle.properties
@@ -2,3 +2,6 @@ ReactNativeViewPager_compileSdkVersion=28
 ReactNativeViewPager_buildToolsVersion=28.0.3
 ReactNativeViewPager_targetSdkVersion=27
 ReactNativeViewPager_minSdkVersion=16
+
+android.enableJetifier=true
+android.useAndroidX=true
--- a/android/src/main/java/com/reactnativecommunity/viewpager/ReactViewPager.java
+++ b/android/src/main/java/com/reactnativecommunity/viewpager/ReactViewPager.java
@@ -7,8 +7,8 @@
 
 package com.reactnativecommunity.viewpager;
 
-import android.support.v4.view.PagerAdapter;
-import android.support.v4.view.ViewPager;
+import androidx.viewpager.widget.PagerAdapter;
+import androidx.viewpager.widget.ViewPager;
 import android.view.MotionEvent;
 import android.view.View;
 import android.view.ViewGroup;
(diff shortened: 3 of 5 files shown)


ferrannp:
Thank you very much! I'll take a look today or tomorrow.

ferrannp:
@aarongrider I tried this in a new generated project `react-native init` with `0.59.5`. It does not work (or require some changes in the main project) because the generated project still does not use Android X. I am guessing we need to wait till `0.60.x` ?

aarongrider:
I'm wondering if we should wait to merge this until RN 60 is released with AndroidX support. As far as I understand, AndroidX projects are incompatible with legacy support projects when compiled from source. This library shouldn't be a huge issue for consumers since you distribute compiled aar's, but it's something we should keep in mind.

Krizzu:
I'd now use it now, as RN 0.59 is still on support libs plus most of the libs did not move to `androidx` just yet

grean:
You should provide a version now even if react-native 0.60 is not yet released...
I could use it now because my project depend on it and it's not working anymore since yesterday -> https://github.com/facebook/react-native/issues/25294
Now other libraries like react-native-webview or react-native-camera have already do the job.
You can merge this PR now and provide a new version, if others don't want androidx they keep previous versions.

noomorph:
0.60 is out. Source: [Announcing React Native 0.60](https://facebook.github.io/react-native/blog/2019/07/03/version-60)

isaax2:
https://github.com/kmagiera/react-native-gesture-handler/pull/460#issuecomment-508576959

example/android/build.gradle
```diff
@@ -5,7 +5,7 @@ buildscript {
         minSdkVersion = 16
         compileSdkVersion = 28
         targetSdkVersion = 27
-        supportLibVersion = "28.0.0"
+        supportLibVersion = "1.0.2"
```
ptrocki: this property name is misleading. Maybe it should be like `appcompatVersion` ?

ferrannp:
We want to wait on this till 2.0 is released (so 2.0 is also compatible before RN 0.60.x). Then we can add these changes for 3.0. For now, if you want to use this library in RN 0.60.x just follow the [jetifier instructions](https://github.com/mikehardy/jetifier).

ivan3177:
@aarongrider @ferrannp @troZee any updates on this one? 2.0.0 is out and RN is already 0.61

Jetifier is doing it's job, however I think it's already time to migrate

bhandarijiwan:
@ferrannp can you please merge this and push a new release ?

ferrannp:
@aarongrider hey I am closing this in favour of https://github.com/react-native-community/react-native-viewpager/pull/103. Sorry for that! But we also need the example working if we update to Android X so we want to speed things up. We'll make a new release when that is merged.

@bhandarijiwan till then just use jetifier or run `react-native run-android` which does it for you.

PR #65, opened 2019-10-05
fix: Remove childrenWithOverriddenStyle.

troZee:
# Summary
We don't need to override styles for each page. This should be handled outside the lib. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅   |
| Android |    ✅  |

## Checklist

### DIFF
--- a/js/utils.js
+++ b/js/utils.js
@@ -18,22 +18,6 @@ export const childrenWithOverriddenStyle = (children?: React.Node): Array<React.
     if (!child) {
       return null;
     }
-    const newProps = {
-      ...child.props,
-      style: [
-        child.props.style,
-        {
-          position: 'absolute',
-          left: 0,
-          top: 0,
-          right: 0,
-          bottom: 0,
-          width: undefined,
-          height: undefined,
-        },
-      ],
-      collapsable: false,
-    };
     if (
       child.type &&
       child.type.displayName &&
@@ -45,6 +29,6 @@ export const childrenWithOverriddenStyle = (children?: React.Node): Array<React.
         child.type.displayName,
       );
     }
-    return React.createElement(child.type, newProps);
+    return React.createElement(child.type, child.props);
   });
 };


troZee:
> Are we sure we do not need it ?
I think, that each view should not be absolute. Children view should relay on its parent. 

> This was taken from the core. I am just wondering if this can introduce a breaking change or if this was done for some specific reason.

I was considering a case, when children view would be just only `ImageView`, then it should be wrapped this `View` tag.  In above case, we added ability to use other View Component than `View`, so dev can decide what should be passed. Because I changed behaviour of children view layouting, so probably it can breaks some views.

I add on hold label, because I need to more test it

troZee:
@ferrannp 
I've checked this code 
```
        <ViewPager
          style={styles.viewPager}
          initialPage={0}
          scrollEnabled={this.state.scrollEnabled}
          onPageScroll={this.onPageScroll}
          onPageSelected={this.onPageSelected}
          onPageScrollStateChanged={this.onPageScrollStateChanged}
          pageMargin={10}
          // Lib does not support dynamically orientation change
          orientation="horizontal"
          // Lib does not support dynamically transitionStyle change
          transitionStyle="scroll" 
          ref={this.viewPager}>
            <Image
            source={{uri: 'https://apod.nasa.gov/apod/image/1410/20141008tleBaldridge001h990.jpg'}}
            />
            <Image
            source={{uri: 'https://apod.nasa.gov/apod/image/1410/20141008tleBaldridge001h990.jpg'}}
          />
          <Text>Some text</Text>
        />
        </ViewPager>
```
And for Images it works well, but for text Red Screen appeared. So, it is better to do above checking.

ferrannp:
> So, it is better to do above checking.

What do you mean?

troZee:
I check each React Node, if it conforms to <View />, if not warning will appeared. I removed code, which is responsible for overriding style for each child.

ferrannp:
I guess with this https://github.com/react-native-community/react-native-viewpager/pull/87 we have enough @troZee ?

PR #70, opened 2019-10-09
Android VerticalPager

devvit:
# Summary

VerticalPager for android, same as ios, ideas come from 
[DeskClock](https://android.googlesource.com/platform/packages/apps/DeskClock/+/master/src/com/android/deskclock/VerticalViewPager.java)
[Vertical ViewPager with HorizontalScrollView inside Fragment](https://stackoverflow.com/questions/46828920/vertical-viewpager-with-horizontalscrollview-inside-fragment)

No js codes changed, keep it clean.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java
+++ b/android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java
@@ -6,7 +6,7 @@ import android.view.GestureDetector;
 import android.view.MotionEvent;
 import android.view.View;
 
-import androidx.viewpager.widget.ViewPager;
+import android.support.v4.view.ViewPager;
 
 // Vertical ViewPager implement, original code from
 // https://android.googlesource.com/platform/packages/apps/DeskClock/+/master/src/com/android/deskclock/VerticalViewPager.java
--- a/android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java
+++ b/android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java
@@ -15,26 +15,22 @@ public class VerticalViewPager extends ViewPager {
     private GestureDetector mGestureDetector;
 
     public VerticalViewPager(Context context) {
-        this(context, null);
-    }
-
-    public VerticalViewPager(Context context, AttributeSet attrs) {
-        super(context, attrs);
-        setOrientation(false);
+        super(context);
     }
 
     public void setOrientation(boolean vertical) {
         mVertical = vertical;
+        if (!mVertical) return;
 
         // Make page transit vertical
-        setPageTransformer(true, mVertical ? new VerticalPageTransformer() : null);
+        setPageTransformer(true, new VerticalPageTransformer());
 
         // Nested scroll issue, follow the link
         // https://stackoverflow.com/questions/46828920/vertical-viewpager-with-horizontalscrollview-inside-fragment
         mGestureDetector = new GestureDetector(getContext(), new GestureDetector.SimpleOnGestureListener() {
             @Override
             public boolean onScroll(MotionEvent e1, MotionEvent e2, float distanceX, float distanceY) {
-                return mVertical ? Math.abs(distanceY) > Math.abs(distanceX) : Math.abs(distanceY) < Math.abs(distanceX);
+                return Math.abs(distanceY) > Math.abs(distanceX);
             }
         });
     }
@@ -57,12 +53,14 @@ public class VerticalViewPager extends ViewPager {
 
     @Override
     public boolean onInterceptTouchEvent(MotionEvent ev) {
-        Boolean result = super.onInterceptTouchEvent(flipXY(ev));
+        boolean result = super.onInterceptTouchEvent(flipXY(ev));
         // Return MotionEvent to normal
         flipXY(ev);
 
-        if (mGestureDetector.onTouchEvent(ev)) {
-            result = true;
+        if (mVertical) {
+            if (mGestureDetector.onTouchEvent(ev)) {
+                result = true;
+            }
         }
 
         return result;
@@ -70,12 +68,14 @@ public class VerticalViewPager extends ViewPager {
 
     @Override
     public boolean onTouchEvent(MotionEvent ev) {
-        Boolean result = super.onTouchEvent(flipXY(ev));
+        boolean result = super.onTouchEvent(flipXY(ev));
         // Return MotionEvent to normal
         flipXY(ev);
 
-        if (mGestureDetector.onTouchEvent(ev)) {
-            result = true;
+        if (mVertical) {
+            if (mGestureDetector.onTouchEvent(ev)) {
+                result = true;
+            }
         }
 
         return result;
--- a/android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java
+++ b/android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java
@@ -29,9 +29,6 @@ public class VerticalViewPager extends ViewPager {
         // Make page transit vertical
         setPageTransformer(true, mVertical ? new VerticalPageTransformer() : null);
 
-        // Get rid of the overscroll drawing that happens on the left and right (the ripple)
-        setOverScrollMode(View.OVER_SCROLL_NEVER);
-
         // Nested scroll issue, follow the link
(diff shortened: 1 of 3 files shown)


android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java
```diff
@@ -0,0 +1,109 @@
...
+    private boolean mVertical = false;
+    private GestureDetector mGestureDetector;
+
+    public VerticalViewPager(Context context) {
+        this(context, null);
+    }
+
+    public VerticalViewPager(Context context, AttributeSet attrs) {
+        super(context, attrs);
+        setOrientation(false);
```
ferrannp: Do we need to call this here? I think to have horizontal ViewPager to behave as it was, we don't really need to do everything that `setOrientation` method is doing, right?

android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java
```diff
@@ -0,0 +1,109 @@
...
+    }
+
+    public void setOrientation(boolean vertical) {
+        mVertical = vertical;
+
+        // Make page transit vertical
+        setPageTransformer(true, mVertical ? new VerticalPageTransformer() : null);
+
+        // Get rid of the overscroll drawing that happens on the left and right (the ripple)
+        setOverScrollMode(View.OVER_SCROLL_NEVER);
```
ferrannp: This is affecting now the vertical and the horizontal. Can we remove this line? Maybe we can add an extra `prop` if uses wants to remove that overscroll ripple effect.

android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * @return {@code false} since a vertical view pager can never be scrolled horizontally
+     */
+    @Override
+    public boolean canScrollHorizontally(int direction) {
+        return !canScrollVertically(direction);
+    }
+
+    /**
+     * @return {@code true} iff a normal view pager would support horizontal scrolling at this time
```
ferrannp: typo: `if` instead of `iff`.

ferrannp:
This is an awesome PR to support `vertical` orientation in Android. Thank you @devvit! Left you some comments.

android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java
devvit: I dont want to change original ```ReactViewPager.java``` much more, so it's necessary with some hooks, ```setPageTransformer``` and ```GestureDetector``` shoud be initialized.

devvit:
@ferrannp, the PR is still support v4, but #22 will migrate to androidx, I think the next version v3.0 shoud be built with [ViewPager2](https://developer.android.com/reference/androidx/viewpager2/widget/ViewPager2.html), just use v2.0 for legacy project
```
ViewPager2 replaces ViewPager, addressing most of its predecessor’s pain-points, 
including right-to-left layout support, vertical orientation, 
modifiable Fragment collections, etc.
```

ferrannp:
Oh I didn't know about [ViewPager2](https://developer.android.com/reference/androidx/viewpager2/widget/ViewPager2.html)!

> I think the next version v3.0 should be built with ViewPager2, just use v2.0 for legacy project :)

I fully agree with that. I think we should hold on with this PR then, things like forcing `setOverScrollMode` are breaking changes.

devvit:
I have removed ```setOverScrollMode```, before v3.0 release, we have to use v2.0, haha.

android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java
devvit: removed

android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java
devvit: removed

ferrannp:
I'll give it more testing time later and we might be able to make a `2.1.0` before `3.0`.

jeanlauliac:
I'm not familiar with view pagers (yet), but I noticed https://github.com/react-native-community/react-native-viewpager has a "vertical" orientation available. Did we try that one out?

troZee:
I found a bug. Initial Screen is not rendering (You can see white screen at startup).
https://media.giphy.com/media/VHlJ9iegvTnTrknIOu/giphy.gif

devvit:
@troZee, I cannot reproduce, tested example on real devices(android 4.4, android 9) and android 10 emu, but confirm: 
1. Dont switch between ```horizontal``` and ```vertical``` at runtime.
2. Development only, ```Hot Module Reload``` will break some features of ViewPager, it happen randomly, just reload all codes.

ferrannp:
@devvit I am seeing this issue in the example:

a93eb0d7-4c15-472a-adef-6d4570208d89

I can actually see the content rendered only when I tap on the screen.

Btw we released 3.0.0 so we should already be able to use the new ViewPager 2.

ferrannp:
@devvit I have a problem with this pager and tapping the screen. See the gif:

finger-stop

When I stop the pager with the finger (tap), when I release it, it show go to the correct page. In here, I am able to just stop it till I touch it again.

devvit:
@ferrannp, try my fix again plz

ferrannp:
Thanks @devvit, that works!. I still think though that we should leave the `horizontal` option as it was, without custom pager and detectors. Only use this for `vertical` pager.

devvit:
The origin idea copy from google source, we need think about how to organize codes.

devvit:
Oh, I have a idea.

```java
// ReactVerticalViewPager.java
public class ReactVerticalViewPager extends ReactViewPager
```

```java
// ReactViewPagerManager.java
protected ReactViewPager createViewInstance(ThemedReactContext reactContext) {
    return orientation == "vertical" ? new ReactVerticalViewPager(reactContext) : new ReactViewPager(reactContext);
}
```

@ferrannp, what about like this do you think ?

ferrannp:
I like it! But will the casting work? Or should it be like this?

```js
protected ViewPager createViewInstance(ThemedReactContext reactContext) {
    return orientation == "vertical" ? new ReactVerticalViewPager(reactContext) : new ReactViewPager(reactContext);
}
```

devvit:
@ferrannp,
Bad news: my above solution failed, ```createViewInstance``` cannot get ```props``` before th instance created.
Good news: the latest commit removed effects on ```horizontal```.

troZee:
@devvit good job. I really appreciate, what you have done. I have tested your changes and it LGTM, but
I would like to wait for @ferrannp . He needs to test it more deeply :D

android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java
```diff
@@ -0,0 +1,117 @@
...
+    private boolean mVertical = false;
+    private GestureDetector mGestureDetector;
+
+    public VerticalViewPager(Context context) {
+        super(context);
+    }
+
+    public void setOrientation(boolean vertical) {
+        mVertical = vertical;
+        if (!mVertical) return;
```
ferrannp: The thing with this is that if we change the prop dynamically (something I tried and it works) we can go from vertical to horizontal and the new gestures will be there. I think we can do:

```js
if (!mVertical) {
  setPageTransformer(null);
  mGestureDetector = null;
  return;
}

Wdyt @devvit ?

**Comment by ferrannp** [comment]
What I am going to do is merge this, state in the docs that `orientation` prop does not work dynamically till we might find a solution for it.

Thank you @devvit !

File: android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java (hunk 4 above)
Comment (devvit) [reply android/src/main/java/com/reactnativecommunity/viewpager/VerticalViewPager.java]: I seen this
```javascript
// example/ViewPagerExample.js
pageMargin={10}
// Lib does not support dynamically orientation change
orientation="horizontal"
```

I'm not sure if ios switch orientation dynamically,  the scene is rare. Maybe we can add the feature after using ViewPager2.

PR #71, opened 2019-10-09
Fixes tapping on UIPageControl to cycle through pages

nkhoshini:
# Summary

* The UIPageControl element added to the view hierarchy manually doesn't cycle through views when tapped. This is because the event has to be added manually. The built in UIPageControl element from UIPageViewController does this automatically, however that's not being used in this project.
* The fix adds a control event hook UIPageControl when the value changes. This is triggered when the user taps it. 

## Test Plan

The current behavior is as such:
react-viewpager-pagecontrol-current

With the fix:
react-viewpager-pagecontrol-fixed

- Add an set of views to **react-native-viewpager**
- Set **showPageIndicator** to **true**
- Tap the UIPageControl itself, observe that the dots update but the views do not

- Add an set of views to **react-native-viewpager**
- Set **showPageIndicator** to **true**
- Tap the UIPageControl itself, behavior works now as expected

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    N/A    |

## Checklist

### DIFF
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -483,7 +483,16 @@ willTransitionToViewControllers:
     pageControl.tintColor = UIColor.blackColor;
     pageControl.pageIndicatorTintColor = UIColor.whiteColor;
     pageControl.currentPageIndicatorTintColor = UIColor.blackColor;
+    [pageControl addTarget:self
+                    action:@selector(pageControlValueChanged:)
+          forControlEvents:UIControlEventValueChanged];
+
     return pageControl;
+}
+- (void)pageControlValueChanged:(UIPageControl *)sender {
+    if (_reactPageIndicatorView.currentPage != _currentIndex) {
+        [self goTo:@(_reactPageIndicatorView.currentPage) animated:YES];
+    }
 }
 
 #pragma mark - UIScrollViewDelegate


dustfire:
Rock star! 💯

nkhoshini:
Thanks @troZee! I'll add code to ensure taps on dots work properly as well 😃

nkhoshini:
@troZee So digging through this some more, the solutions you've provided work for the tapable region outside the dots themselves. I have in the commit setup the event to trigger off of **UIControlEventValueChanged** as the folks in StackOverflow suggested so the PR is doing the same thing. 

Tapping on the dot itself is tricky. iOS renders them as individual views in the hierarchy and there is no event triggered when tapping on it directly:

image

Oddly enough, Apple's own HI guidelines mention UIPageControl components are not meant to updated when the dot itself is tapped:

https://developer.apple.com/design/human-interface-guidelines/ios/controls/page-controls/

> Users can tap the leading or trailing edge of a page control to visit the next or previous page, but they can't tap a specific dot to go to a specific page. Navigation always occurs sequentially, usually by swiping the page to one side

However this is contradictory to how it works on the home screen (Springboard) 🤷‍♂. At any rate, it looks like short of digging into the view hierarchy of the page controller component, getting the tap to work on the individual dot maybe difficult to do cleanly:

https://stackoverflow.com/questions/31818218/uipagecontrol-select-dot-and-go-to-that-page

troZee:
Thank you @nkhoshini for deep investigation. I really love it <3. Now I understand the problem. 

I am so surprised about this thing:  
> Users can tap the leading or trailing edge of a page control to visit the next or previous page, but they can't tap a specific dot to go to a specific page. Navigation always occurs sequentially, usually by swiping the page to one side

Just 🍎things.

PR #76, opened 2019-10-15
feat: add eslint

troZee:
# Summary
Add ESlint to project

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,9 @@
   "scripts": {
     "start": "node node_modules/react-native/local-cli/cli.js start",
     "test": "echo \"Error: no test specified\" && exit 1",
-    "flow": "flow"
+    "flow": "flow",
+    "lint": "eslint .",
+    "fix": "eslint . --fix --ext .js"
   },
   "keywords": [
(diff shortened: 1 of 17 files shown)


dratwas:
Hey, @troZee I had to remove `resource_class` since we have a free plan. PR looks good but please add a job to CI where you run lint check :)

package.json
```diff
@@ -7,7 +7,9 @@
   "scripts": {
     "start": "node node_modules/react-native/local-cli/cli.js start",
     "test": "echo \"Error: no test specified\" && exit 1",
-    "flow": "flow"
+    "flow": "flow",
+    "lint": "eslint .",
+    "fix": "eslint . --fix --ext .js"
```
ferrannp: I find this redundant. I'd say `yarn lint --fix` is enough. Or at least the script to be `lint:fix`. Just `fix` is very generic.

dratwas: Yeah I would leave the script for fixing but as @ferrannp said `fix` is too generic. Let's use `lint:fix`

ferrannp:
Thank you!

PR #86, opened 2019-10-28
fix: Xcode 11 error

troZee:
# Summary
fix: add patch to react package

Close https://github.com/react-native-community/react-native-viewpager/issues/85

## Test Plan
Just run the app

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,8 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "flow": "flow",
     "lint": "eslint .",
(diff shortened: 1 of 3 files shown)


ferrannp:
@troZee I am pretty sure we do not want this in here. If somebody is using `react-native 0.58.4` and they need to patch it, it is something they need to do in their app. The fact that you patch it here won't affect the app that is consuming this library.

PR #102, opened 2019-11-12
fix: scrollEnabled prop, when comp is loaded dynamically

troZee:
# Summary

Close https://github.com/react-native-community/react-native-viewpager/issues/99
Close https://github.com/react-native-community/react-native-viewpager/issues/96
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ❌     |

## Checklist


fcostarelli-tbox:
This fixed it for me, but only if `transitionStyle={'scroll'}`. When set to `curl` I can still scroll even with `scrollEnabled={false}`. Any suggestion on how this could be fixed?

PR #114, opened 2019-11-26
Revert "Add overdrag property on iOS"

osdnk:
Reverts react-native-community/react-native-viewpager#107

It sometimes breaks on iOS. Pager stucks in the middle of the animation. Until I figure out what's wrong, I need to revert this

image


Aryk:
Any updates when this might make it back in? I can't really use it with the stretching/bouncing on drag.

PR #119, opened 2019-12-03
feat: add imperative setScrollEnabled method

### LINKED ISSUE #112: Be able to change scrollEnabled imperatively
# Feature Request

Right now we can change `scrollEnabled` only via props. I have a use case when I run an animation and I want to disable the pager scroll while the animation is running. 

## Why it is needed

Animations do not behave well (lag) if you change / re-render the component while it is happening.

I tried a solution locally like:

```js
viewPager.current.setScrollEnabled(false);
```

And that just solves my problems. Although the recommended way should be via props just to keep it in sync (if not the developer needs to remember to keep it in sync).

## Possible implementation

Just the same as for example `setPage` to call the native method imperatively.

### Code sample

```js
// Using PanGestureHandler from react-native-gesture-handler
 _onHandlerStateChange = ({nativeEvent}) => {
  if (nativeEvent.oldState === State.ACTIVE) {
    viewPager.current.setScrollEnabled(false);
    Animated.spring(...).start(() => {
      viewPager.current.setScrollEnabled(true);
    })
  }
}
```

### LINKED ISSUE #117: Vertical Android ViewPager not working correctly in certain devices
Swiping for a short distance and fast should change the page. This is working fine in my Pixel 2 (Android 10) and in Android the emulator. However, we tried a couple of devices where is not working correctly:

- Samsung Galaxy Note 9 (Android 9)
- Xiaomi MI MIX 2 (Android 9)

This is the behavior on for the vertical ViewPager:

vertical

Basically the user needs to scroll all the way down or up (manually) to trigger a change of page. With the horizontal ViewPager (recorded on the same phone), it works just as expected:

horizontal

Would love to get some help from you @devvit if you know what might be causing this.

ferrannp:
Fixes #112.

# Summary

Summary is in #112.

## Test Plan

Can be tried using `this.viewPager.current.setScrollEnabled(false)` (or `true`) anywhere in our example. I am not adding this in the `README` as is kind advanced / not recommended.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/ios/ReactViewPagerManager.m
+++ b/ios/ReactViewPagerManager.m
@@ -46,19 +46,6 @@ RCT_EXPORT_VIEW_PROPERTY(onPageScrollStateChanged, RCTDirectEventBlock)
         [view shouldScroll:enabled];
     }];
 }
-: (nonnull NSNumber *)reactTag enabled
-: (nonnull BOOL *)enabled {
-    [self.bridge.uiManager addUIBlock:^(
-                                        RCTUIManager *uiManager,
-                                        NSDictionary<NSNumber *, UIView *> *viewRegistry) {
-        ReactNativePageView *view = (ReactNativePageView *)viewRegistry[reactTag];
-        if (!view || ![view isKindOfClass:[ReactNativePageView class]]) {
-            RCTLogError(@"Cannot find ReactNativePageView with tag #%@", reactTag);
-            return;
-        }
-        [view shouldScroll:enabled];
-    }];
-}
 
 RCT_EXPORT_METHOD(setPage
                   : (nonnull NSNumber *)reactTag index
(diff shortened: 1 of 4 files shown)


ios/ReactViewPagerManager.m
```diff
@@ -32,6 +32,21 @@ - (void) goToPage
     }];
 }
 
+- (void) changeScrollEnabled
```
troZee:
```suggestion
- (void) changeScrollEnabled
: (nonnull NSNumber *)reactTag enabled
: (BOOL)enabled {
    [self.bridge.uiManager addUIBlock:^(
                                        RCTUIManager *uiManager,
                                        NSDictionary<NSNumber *, UIView *> *viewRegistry) {
        ReactNativePageView *view = (ReactNativePageView *)viewRegistry[reactTag];
        if (!view || ![view isKindOfClass:[ReactNativePageView class]]) {
            RCTLogError(@"Cannot find ReactNativePageView with tag #%@", reactTag);
            return;
        }
        [view shouldScroll:enabled];
    }];
}
```

ios/ReactViewPagerManager.m
```diff
@@ -44,6 +59,13 @@ - (void) goToPage
     [self goToPage:reactTag index:index animated:false];
 }
 
+RCT_EXPORT_METHOD(setScrollEnabled
```
troZee:
```suggestion
RCT_EXPORT_METHOD(setScrollEnabled
                  : (nonnull NSNumber *)reactTag enabled
                  : (BOOL) enabled) {
    [self changeScrollEnabled:reactTag enabled:enabled];
}
```

PR #131, opened 2019-12-12
Allow setting tintColor of iOS PageView component

baugarten:
# Summary
* Allows configuration of the tintColor of the PageView component on iOS
* See https://github.com/react-native-community/react-native-viewpager/issues/98

## Test Plan

I verified it works by installing it locally on a project that I'm working on. Included are screenshots of the configuration and end result

Simulator Screen Shot - iPhone 11 - 2019-12-12 at 10 16 25
Screen Shot 2019-12-12 at 10 16 43 AM

An embedded ViewPager in a react-native app

Include the properties 
```
currentPageIndicatorTintColor={'#00FFFF'}
pageIndicatorTintColor={'#FF0000'}
```
on the view pager, reload, verify that the active and inactive tint colors have been changed (iOS only)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

  - I don't see a `CHANGELOG.md` in the repository. I'm happy to add this to the appropriate place.
- [x] I added a sample use of the API in the example project (`example/App.js`)

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,8 @@ const styles = StyleSheet.create({
 |`orientation: Orientation`|Set `horizontal` or `vertical` scrolling orientation (it does **not** work dynamically)|both
 |`transitionStyle: TransitionStyle`|Use `scroll` or `curl` to change transition style (it does **not** work dynamically)|iOS
 |`showPageIndicator: boolean`|Shows the dots indicator at the bottom of the view|iOS
+|`currentPageIndicatorTintColor: string`|A hexadecimal representation of the color for the active tab of the page indicator|iOS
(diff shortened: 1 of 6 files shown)


troZee:
Hey 👋  Thank you for your contribution. From my POV it is better to have one property eg. `pageIndicatorProps` (no `xxxStyle` because it can be confused) and this property would have below shape: 

```
{
currentTintColor? : string,
tintColor?: string,
.
.
.
// some others props 
}
```

wilsolutions:
I've just started looking into this project and I cannot see the inactive dots because of the they are white and the App I'm testing has a white bg as well.
This PR seems to solve the issue but has been parked here for awhile now...
Any thoughts?

PR #138, opened 2020-01-29
update index.d.ts to match flowtype

Naturalclar:
# Summary

fixes `index.d.ts` and removes `any` from the type.

The original `index.d.ts` had `ViewPager` extend `ViewProps`, but in the flow type, defined in https://github.com/react-native-community/react-native-viewpager/blob/master/js/types.js, it does not actually extend `ViewProps`, it only uses `ViewProps` to get the required type for `onMoveShouldSetResponderCapture`

## Checklist

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -14,7 +14,7 @@ export interface PageScrollStateChangedEvent {
     pageScrollState: 'idle' | 'dragging' | 'settling';
 }
 
-export interface ViewPagerProps extends ReactNative.ViewProps {
+export interface ViewPagerProps {
     /**
      * Index of initial page that should be selected. Use `setPage` method to
      * update the page, and `onPageSelected` to monitor page changes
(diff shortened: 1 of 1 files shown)


typings/index.d.ts
```diff
@@ -69,7 +69,20 @@ export interface ViewPagerProps extends ReactNative.ViewProps {
      */
     pageMargin?: number;
 
-    onMoveShouldSetResponderCapture?: (event: any) => boolean;
+    style?: ReactNative.StyleProp<ReactNative.ViewStyle>
+
+    children?: React.ReactChildren
+
```
antonycohen, line 75: I've got this, since this update
"Type 'Element[]' is not assignable to type 'ReactChildren'."

troZee: can you try with ReactNode ? 

I found it here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L1333

antonycohen: No problem with ReactNode, sry for delay, didn't see your comment.

PR #166, opened 2020-05-04
fix: Set type on children as ReactNode in typings

### LINKED ISSUE #165: [ts] Wrong children type
I've got this, since this update
"Type 'Element[]' is not assignable to type 'ReactChildren'."

_Originally posted by @antonycohen in https://github.com/react-native-community/react-native-viewpager/pull/138_

mikaelbr:
# Summary

With the latest changes in typings (e12c183636f3b69c8e2408ccf82318b133034e74) as an attempt to match Flow and TypeScript, `ReactChildren` was used as children type. This is not the correct one, AFAIK. The typical children type is `ReactNode` which is defined as:

```ts
type ReactNode = ReactChild | ReactFragment | ReactPortal | boolean | null | undefined;
```

## Test Plan

This is a change in the typings and there are no proper test cases for this in the repo, but this has been tested in an actual project where the issue with mistypes occurred.

## Checklist

---

This PR Fixes #165

### DIFF
--- a/typings/index.d.ts
+++ b/typings/index.d.ts
@@ -69,9 +69,9 @@ export interface ViewPagerProps {
      */
     pageMargin?: number;
 
-    style?: ReactNative.StyleProp<ReactNative.ViewStyle>
+    style?: ReactNative.StyleProp<ReactNative.ViewStyle>;
 
-    children?: React.ReactChildren
+    children: React.ReactNode;
 
     /**
      * If a parent `View` wants to prevent a child `View` from becoming responder


typings/index.d.ts
```diff
@@ -69,9 +69,9 @@ export interface ViewPagerProps {
      */
     pageMargin?: number;
 
-    style?: ReactNative.StyleProp<ReactNative.ViewStyle>
+    style?: ReactNative.StyleProp<ReactNative.ViewStyle>;
 
-    children?: React.ReactChildren
+    children: React.ReactNode;
```
troZee, line 74: Could you add `?` to have consistency with flow types?  :
https://github.com/react-native-community/react-native-viewpager/blob/master/js/types.js#L102

mikaelbr: I don't think that is needed as `ReactNode` is union type with case `undefined` itself (see my description of the PR)

satya164: Even with Flow, shouldn't children be required?

mikaelbr: I don't think it is unreasonable to have no children? I don't have a particular use case in mind other than for instance populating items dynamically and having transitions/animations for inserting new nodes. As far as I could see in the JS-part of it supports null/undefined cases at least.

satya164: even if you're populating data from an API, you'd likely be passing an empty array rather than null or undefined, though same argument can be applied to any component. regardless, I'm not talking about passing null or undefined. the question mark marks the prop as optional, which is different from requiring to pass undefined/null explicitly. optional children definitely not correct usage in this case. so the flow types shouldn't have ? either.

PR #171, opened 2020-05-10
perf: implement coalescing for onPageScroll on iOS

Almouro:
# Summary

Found inspiration in https://github.com/software-mansion/react-native-gesture-handler/pull/997:
This pull request implements event coalescing on `onPageScroll` event on iOS. 

Event coalescing is a built-in mechanism from React Native's event-emitter that merges several events together when events are generated faster on the native thread than they can be handled on the Javascript thread. This prevents accumulating lag when JS thread is under heavy load (or on low-end devices).

NB: coalescing is [the default on Android](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/uimanager/events/Event.java#L56) so no changes on Android were required
This PR basically implements the same behavior on iOS.

The code below uses the fibonacci function to simulate a busy JS thread
```js
function fibo(n: number): number {
  return n <= 2 ? n : fibo(n - 1) + fibo(n - 2);
}

const AnimatedViewPager = Animated.createAnimatedComponent(ViewPager);

const App = () => {
  const [text, setText] = useState(0);

  const scrollOffset = useRef(new Animated.Value(0)).current;
  const position = useRef(new Animated.Value(0)).current;

  return (
    <>
      <Text style={styles.text}>{text}</Text>
      <AnimatedViewPager
        style={styles.viewPager}
        initialPage={0}
        onPageScroll={Animated.event([{ nativeEvent: { offset: scrollOffset, position } }], {
          listener: ({ nativeEvent: { offset, position } }) => {
            fibo(35);
            setText(offset + position);
          },
          useNativeDriver: true,
        })}
      >
        <View key="1" style={{ backgroundColor: "yellow" }}>
          <Text style={{ fontSize: 30 }}>First page</Text>
        </View>
        <View key="2" style={{ backgroundColor: "blue" }}>
          <Text style={{ fontSize: 30 }}>Second page</Text>
        </View>
      </AnimatedViewPager>
    </>
  );
};
```

Here are GIFs demonstrating before and after the fix:

| Before Fix      | After Fix |
| ------- | :---------: |
|pager-beforecoalescing|pager-aftercoalescing|

Without event coalescing, the text label continues to be updated long after the gesture is actually ended. With event coalescing, it stops updating as soon as the scrolling ends, since there is no event accumulation in the native thread's event-queue.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅ **(no changes required)**     |

## Checklist

### DIFF
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -8,8 +8,7 @@
 
 - (instancetype) initWithReactTag:(NSNumber *)reactTag
                          position:(NSNumber *)position
-                           offset:(NSNumber *)offset
-                    coalescingKey:(uint16_t)coalescingKey;
+                           offset:(NSNumber *)offset;
 
 @end
 
@@ -17,7 +16,6 @@
 {
     NSNumber* _position;
     NSNumber* _offset;
-    uint16_t _coalescingKey;
 }
 
 @synthesize viewTag = _viewTag;
@@ -28,8 +26,7 @@
 
 - (instancetype) initWithReactTag:(NSNumber *)reactTag
                          position:(NSNumber *)position
-                           offset:(NSNumber *)offset
-                    coalescingKey:(uint16_t)coalescingKey;
+                           offset:(NSNumber *)offset;
 {
     RCTAssertParam(reactTag);
     
@@ -37,7 +34,6 @@
         _viewTag = reactTag;
         _position = position;
         _offset = offset;
-        _coalescingKey = coalescingKey;
     }
     return self;
 }
@@ -45,13 +41,13 @@
 RCT_NOT_IMPLEMENTED(- (instancetype)init)
 - (uint16_t)coalescingKey
 {
-    return _coalescingKey;
+    return 0;
 }
 
 
 - (BOOL)canCoalesce
 {
-    return NO;
+    return YES;
 }
 
 + (NSString *)moduleDotMethod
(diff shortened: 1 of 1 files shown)


troZee:
Thank you @Almouro for 2 great PR ❤️ 
Could you resolve conflicts ?

Almouro:
Thanks @troZee, the pleasure was mine 🎉 
Solved the conflict :)

troZee:
For iOS, it works like a charm. 

It does not work on Android
coalescingKey

When I remove `fibo(35)`, it work on Android, too. 
Would like to fix android ?

troZee:
Probably, you need to add below chunk https://github.com/software-mansion/react-native-gesture-handler/pull/997/files#diff-e6b4a0b8d8f24f0ab4b92f46c7c47355R34 

to fix android

Almouro:
@troZee my guess is that `fibo(35)` is too much to calculate for the emulator, so the JS thread is completely locked and can never reach the `setText` line.
(you can enable the perf monitor to double check that)

Here's what I've tried to make sure coalescing is indeed working on Android:

In `android/src/main/java/com/reactnativecommunity/viewpager/PageScrollEvent.java`, add those lines in the class to explicitly disable coalescing:

```java
@Override
public boolean canCoalesce(){
  return false;
}
```

Results on Samsung S10 Android 10, **using fibo(32)** instead of 35

| Current behavior     | Coalescing explicitly disabled with the above |
| ------- | :---------: |
| android-with-coalescing |android-without-coalescing|

In conclusion, event coalescing is the default for events on Android already as we can see [here](https://github.com/facebook/react-native/blob/master/ReactAndroid/src/main/java/com/facebook/react/uimanager/events/Event.java#L56) :)

PR #184, opened 2020-06-18
fix(ios): hide/show indicator once orientation changed

troZee:
# Summary

When orientation changed, page indicator did not behave in correct way. Instead of be hidden, it was displayed on the screen. Initially in portrait mode indicator was not displayed, but in landscape mode, when orientation changed, indicator was displayed. It should be hidden all the time, unless `showPageIndicator` is set to `true`. 

I noticed, in curl mode `setHidden` does not work. I could find any reliable solution for it and I hacked a little bit VP. 

Closes https://github.com/react-native-community/react-native-viewpager/issues/182

## Test Plan

* set ` transitionStyle="curl"`
* rotate the screen 
* check, if dots are hidden 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -228,6 +228,7 @@ RCT_NOT_IMPLEMENTED(- (instancetype)init)
         [self shouldScroll:_scrollEnabled];
         //Below line fix bug, where the view does not update after orientation changed.
         [self goTo:@(_currentIndex) animated:NO];
+        [self shouldShowPageIndicator:_showPageIndicator];
     } else {
         [self embed];
     }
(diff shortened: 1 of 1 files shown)


ios/ReactNativePageView.m
```diff
@@ -472,6 +473,17 @@ - (void)shouldShowPageIndicator:(BOOL)showPageIndicator {
     if (_reactPageIndicatorView){
         _reactPageIndicatorView.hidden = !showPageIndicator;
     }
+    
+    // https://github.com/react-native-community/react-native-viewpager/issues/182
+    if (_transitionStyle == UIPageViewControllerTransitionStylePageCurl) {
+        NSArray *subviews = self.reactPageViewController.view.subviews;
+        for (int i=0; i<[subviews count]; i++) {
```
michalchudziak, line 480: You can use
```objc
for (id subview in subviews) {
...
}
```

if your min target is iOS 10 :)

ios/ReactNativePageView.m
```diff
@@ -472,6 +473,17 @@ - (void)shouldShowPageIndicator:(BOOL)showPageIndicator {
...
+            if ([[subviews objectAtIndex:i] isKindOfClass:[UIPageControl class]]) {
```
mdjastrzebski, line 481: You can use `subviews[i]` instead of `[subviews objectAtIndex:i]` (the same also below).

troZee: I used https://github.com/react-native-community/react-native-viewpager/pull/184#discussion_r442012212
hence I don't need to use index anymore :D

troZee:
I would like to dig more to find root cause of this issue, hence I hold this PR.

cc @dratwas

robban:
I can confirm that this fixes the issue in my app, thank you 👍

troZee:
@robban thank you for checking this PR. You can use it as temporary workaround using patch-package. I would like to dig more to find better fix.

troZee:
it has been fixed here: https://github.com/react-native-community/react-native-viewpager/pull/189

PR #189, opened 2020-07-01
Improve iOS code to use datasource behaviour from UIPageViewController

krystiansliwa:
# Summary

After reviewing iOS code I realized that it can be simplified and optimized a bit. 

This PR changes a way of handling controllers in native code. 
Previously each view had associated controller that was stored in the memory and each time updated with view if needed. Now controller is created or reused on demand for a particular view. 
Also current implementation uses the same method of creating/reusing controller for gesture-based and goTo navigation inside swiper which simplify code.

Some of the previous issues like **ViewControllerHierarchyInconsistency** should be resolved now.

For storing controllers I decide to use **NSHashTable** with "weak option" so we are sure that controllers are not retained.

To improve readability I extracted **Events** classes to corresponding files and moved some methods/properties declarations  from .h to .m files.

## Test Plan

Just check if it works. There was no API changes, only implementation was changed.

## Checklist


ios/ReactNativePageView.m
```diff
@@ -218,307 +40,299 @@ - (instancetype)initWithEventDispatcher:(RCTEventDispatcher *)eventDispatcher {
...
-    
-    index++;
-    
-    if (index == [_childrenViewControllers count]) {
-        return nil;
+    if (!controllerToDisplay) {
+         controllerToDisplay = [[UIViewController alloc] initWithView:viewToDisplay];
     }
-    return [_childrenViewControllers objectAtIndex:index];
+    [self.cachedControllers addObject:controllerToDisplay];
```
troZee, line 224: should we always add controller to `cachedControllers`
IMO it should be only, when `controllerToDisplay` is null

troZee:
@krystiansliwa 
Thank you for veeery big refactor. I really appreciate it. I left only one comment. Once merge conflict resolved, I will merge it. 

Good job :D

ios/ReactNativePageView.m
krystiansliwa: I am doing that just to be sure to have all controllers cached. They are not retained so no risk

PR #201, opened 2020-07-23
chore: configure release-it

troZee:
# Summary

Add `release-it` to automate release process. 

## Test Plan

run `yarn release-it --dry-run` in project dir

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@ version: 2.1
 
 # Make sure you use the latest version of the Orb!
 orbs:
-  rn: react-native-community/react-native@2.0.1
+  rn: react-native-community/react-native@4.4.2
 
 # Custom jobs which are not part of the Orb
 jobs:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -56,5 +56,6 @@ workflows:
           project_path: example/ios/ViewPagerExample.xcodeproj
           build_configuration: Debug
           scheme: ViewPagerExample
+          device: 'iPhone 11'
           requires:
             - analyse_js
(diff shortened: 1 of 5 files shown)


example/android/app/build.gradle
```diff
@@ -210,6 +210,18 @@ dependencies {
         implementation jscFlavor
     }
 }
+task downloadDependencies() {
+    description 'Download all dependencies to the Gradle cache'
+    doLast {
+        configurations.findAll().each { config ->
+            if (config.name.contains("minReactNative") && config.canBeResolved) {
+                print config.name
```
Krizzu, line 218: do you need `print` here?

package.json
```diff
@@ -11,7 +11,8 @@
     "test": "echo \"Error: no test specified\" && exit 1",
     "flow": "flow",
     "lint": "eslint .",
-    "lint:fix": "eslint . --fix --ext .js"
+    "lint:fix": "eslint . --fix --ext .js",
```
Krizzu, line 14: I think `.js` is default, so no need to specify

.circleci/config.yml
```diff
@@ -1,183 +1,60 @@
...
+      - rn/android_build:
+          name: build_android
+          project_path: "example/android"
+          build_type: debug
           requires:
-            - linux-checkout
-      # - jest:
+            - analyse_js
+
+      # Build the iOS app in release mode and do not run tests
```
wojteg1337, line 51: Do you need it?

troZee: I would like to check, if iOS building correctly, but probably, this repo has no permission to macOS container. Now I have no time to check it. I would like to do it later.

PR #203, opened 2020-07-25
fix: Android returning blank view when collection mutates

gusparis:
# Summary
In Android, when component collection sent to the ViewPager mutates for some reason (eg. a list of questions in which the next question depends on the previous answer, see example below), the `ViewPagerFragment` will return an empty view.

## Test Plan

After a little debugging, I realized the observers weren't being notified of the change through `notifyDataSetChanged()`. As stated in [Android's documentation](https://developer.android.com/reference/kotlin/androidx/recyclerview/widget/RecyclerView.Adapter#notifydatasetchanged), it is more efficient to specify the event change.

> If you are writing an adapter it will always be more efficient to use the more specific change events if you can. Rely on notifyDataSetChanged() as a last resort.

## Compatibility

Impacts on Android only.


troZee:
Hey @gusparis 
Thank you for your contribution ❤️. Definitely it should be `notifyItemChanged`. I have checked your changes and It looks good to me.

Issue #211, opened 2020-07-30
Removing a page not working on Android 


mrbrentkelly:
# Bug report

## Summary

I am unable to remove pages in the ViewPager other than the last page. 

Say I have a ViewPager with 3 pages. If I remove the last page (page 3) it works perfect. But if I remove page 2 I am left with a blank screen. 

**This is it working when deleting the last page**

<img src="https://user-images.githubusercontent.com/5814579/88928894-450f7600-d271-11ea-9a06-05475d849eb3.gif" width="200"/>

**And this is it broken when deleting page 2 (of 3)**

<img src="https://user-images.githubusercontent.com/5814579/88928907-48a2fd00-d271-11ea-8382-4c5004a89931.gif" width="200"/>

> Note this is working fine on iOS

## Environment info

`react-native info` output:

```bash
System:
    OS: macOS 10.15.6
    CPU: (12) x64 Intel(R) Core(TM) i7-8750H CPU @ 2.20GHz
    Memory: 396.14 MB / 32.00 GB
    Shell: 3.2.57 - /bin/bash
  Binaries:
    Node: 12.16.3 - ~/.nvm/versions/node/v12.16.3/bin/node
    Yarn: 1.22.4 - ~/dev/hudl/temp/ViewPagerBug/node_modules/.bin/yarn
    npm: 6.14.4 - ~/.nvm/versions/node/v12.16.3/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  Managers:
    CocoaPods: 1.9.3 - /Users/brent.kelly/.rvm/gems/ruby-2.6.3/bin/pod
  SDKs:
    iOS SDK:
      Platforms: iOS 13.6, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK:
      API Levels: 23, 26, 27, 28, 29
      Build Tools: 28.0.2, 28.0.3, 29.0.0, 29.0.1, 29.0.2, 29.0.3, 30.0.0
[...]

PR #214, opened 2020-08-06
fix (android): #211 pages can now be removed on Android

mrbrentkelly:
# Summary

This PR fixes https://github.com/react-native-community/react-native-viewpager/issues/211 which was a bug preventing pages from being deleted on Android.

The solution was to override `getItemId()` in `FragmentAdapter.java` to provide a unique ID (as opposed to the item position, which was the default). See https://stackoverflow.com/a/26944013/2904572

I also had to update the `createFragment` method `FragmentAdapter` to return a new Fragment on each call. Since we keep the internal views around anyway I think this is fine (and a new fragment is always expected as per the Android docs as far as I can tell). 

> _"Make sure that your new createFragment() method always supplies a new fragment instance each time the function is called instead of reusing instances"_
https://developer.android.com/training/animation/vp2-migration#adapter-methods

## Test Plan

Used the repro steps in https://github.com/react-native-community/react-native-viewpager/issues/211. Applied my changes, re-ran the app and the issue was gone.

Create a ViewPager with 3 (or more) pages. Delete a page in-between the first and last page.

See https://github.com/react-native-community/react-native-viewpager/issues/211 for steps to reproduce

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativecommunity/viewpager/FragmentAdapter.java
+++ b/android/src/main/java/com/reactnativecommunity/viewpager/FragmentAdapter.java
@@ -6,12 +6,9 @@ import androidx.annotation.NonNull;
 import androidx.fragment.app.Fragment;
 import androidx.fragment.app.FragmentActivity;
 
-
 import com.reactnative.community.viewpager2.adapter.FragmentStateAdapter;
 
 import java.util.ArrayList;
-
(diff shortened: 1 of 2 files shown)


troZee:
Hey @mrbrentkelly , 
I am happy, you resolved the problem and spent some time to fix it. 
I have tested it on the simulator and LGTM.

PR #223, opened 2020-08-28
fix(android): disable touchable view, when user swipes

troZee:
# Summary

close https://github.com/react-native-community/react-native-viewpager/issues/164

## Test Plan

* create list with touchable items 
* check, if touchable items will trigger onPress() when user scrolls

You can checkout on `164` branch and check, if it works.  

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅    |

## Checklist

### DIFF
--- a/js/ViewPager.js
+++ b/js/ViewPager.js
@@ -164,8 +164,6 @@ class ViewPager extends React.Component<ViewPagerProps> {
    * See https://github.com/react-native-community/react-native-viewpager/issues/164
    */
   _onMoveShouldSetResponderCapture = () => true;
-    return true;
-  };
 
   render() {
     return (
--- a/js/ViewPager.js
+++ b/js/ViewPager.js
@@ -83,8 +83,6 @@ function getViewManagerConfig(viewManagerName) {
  */
 
 class ViewPager extends React.Component<ViewPagerProps> {
-  isScrolling = false;
-
   componentDidMount() {
     // On iOS we do it directly on the native side
     if (Platform.OS === 'android') {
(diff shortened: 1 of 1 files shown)


js/ViewPager.js
```diff
@@ -162,11 +159,12 @@ class ViewPager extends React.Component<ViewPagerProps> {
     );
   };
 
+  /**
+   * `ViewPager` wants to prevent a child `View` from becoming responder on a move
+   * See https://github.com/react-native-community/react-native-viewpager/issues/164
+   */
   _onMoveShouldSetResponderCapture = () => {
```
Krizzu, line 166:
```suggestion
  _onMoveShouldSetResponderCapture = () => true;
```

small nit

js/ViewPager.js
```diff
@@ -163,7 +163,7 @@ class ViewPager extends React.Component<ViewPagerProps> {
    * `ViewPager` wants to prevent a child `View` from becoming responder on a move
    * See https://github.com/react-native-community/react-native-viewpager/issues/164
    */
-  _onMoveShouldSetResponderCapture = () => {
+  _onMoveShouldSetResponderCapture = () => true;
     return true;
```
Krizzu: need to remove this now 🕺

luatnd: I cannot touch anything if it returns `true`
Why always `true` instead of `this.isScrolling`?

luatnd: If I return this.isScrolling => Can scroll but sometimes I have accidental touches.
If I return false => It always causes accidental touches.

troZee:
This change does not fix the problem

PR #242, opened 2020-11-05
Fix Android multiple pager instances

alpha0010:
# Summary

When there are multiple pagers, especially if some are dynamically created/removed, pages can disappear.  This only occurs on Android due to clearing the entire static array.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅ (already works)     |
| Android |    ✅     |

## Checklist


troZee:
https://github.com/callstack/react-native-viewpager/releases/tag/v5.0.2

PR #245, opened 2020-11-24
Add offscreen page limit

osdnk:
# Summary
I need this so deeply https://developer.android.com/reference/androidx/viewpager2/widget/ViewPager2#setOffscreenPageLimit(int)
Basically, I have a ton of things rendered on pages and do not want to optimize it, because it leads to some minor glitches 

Before: 
https://streamable.com/psex6e

After:
https://streamable.com/nq5yvb

## Test Plan

Doesn't have too much impact unless the app is really packed with components. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist


troZee:
Screen Shot 2020-11-24 at 4 46 19 PM

osdnk:
Oh god, now i see that you reverted to viewpager1. 

So it won't be working. 

What was the motivation beyond reverting this?

troZee:
> Oh god, now i see that you reverted to viewpager1.
You can point to `next` branch 
> What was the motivation beyond reverting this?
Lots issues regarding to touching events.

PR #274, opened 2021-01-13
chore: add prop to disable/enable scroll bounces

### LINKED ISSUE #120: Disable "bouncing" for iOS
Similar to what React Native does with `ScrollView` or `FlatList` when `bounces = false`.

arekkubaczkowski:
# Summary
Closes #120  

It's a missing feature which was reverted in certain version which allows to disable/enable bounce effect while scrolling on iOS.
https://stackoverflow.com/questions/21798218/disable-uipageviewcontroller-bounce


troZee:
https://github.com/callstack/react-native-viewpager/issues/266

Issue #140, opened 2020-01-31
Reimplement iOS VP implementation


troZee:
# Feature Request
While I was implementing VP2 on android, which is based on RecyclerView I noticed, that listView implementation is more performant than current implementation. VP on iOS is based on `UIPageViewController `.  I would like to migrate into CollectionView ([example](https://github.com/WenchaoD/FSPagerView)).

## Why it is needed
This implementation should: 
  - fix lazy loading issue
  - add ability to create a transition between pages (Because UIPageViewController does not support it)
  - enable or disable bouncing on iOS

MrRefactor:
Closing as implemented with 8.0.0

PR #276, opened 2021-01-15
Lazy pager [feedback wanted]

### LINKED ISSUE #104: Dynamic / lazy approach to load pages
# Feature Request

## Why it is needed

```js
const min = 0;

  <View key={item}>
      <Item item={item} />
    ) : null}
  </View>
));

<ViewPager ... >
  {items}
</ViewPager>
```

```js
```

alpha0010:
# Summary

Enables pages to be rendered on demand, instead of all at once.  Important when there are a very large number of pages.

Currently, a basic implementation on Android.  I just wiped the app, and init fresh with `react-native-builder-bob`, so will need to restore/rename some files later.

For iOS, I noticed #140 .  Should this lazy render implementation avoid `UIPageViewController`?

## Test Plan

In progress...

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

Fixes #104


alpha0010:
Any thoughts on the architecture of rendering/passing views?

troZee:
Hey @alpha0010 
Thank for your contribution. As you can see there is lots of changes here, so could you describe it in simple words, how you achieved it ? bc basically native side (android/ios) supports lazy loading. The problem is with react-native itself, bc it loads the child views eagerly, so `didMount` methods is called immediately.

alpha0010:
- JS tells native how many total pages there will be
  - https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/src/index.tsx#L36
  - https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/android/src/main/java/com/reactnativecommunityviewpager/ViewPagerViewManager.kt#L63
- Native lazy renders that many (blank) pages
  - https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/android/src/main/java/com/reactnativecommunityviewpager/ReactViewAdapter.kt#L26
  - https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/android/src/main/java/com/reactnativecommunityviewpager/ReactViewAdapter.kt#L29-L32
  - https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/android/src/main/java/com/reactnativecommunityviewpager/PageFragment.kt#L16
- Native tells js which page is being displayed
  - https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/android/src/main/java/com/reactnativecommunityviewpager/ViewPagerViewManager.kt#L22-L26
  - https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/src/index.tsx#L39
- JS renders the visible page (and *n*-pages before/after)
  - https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/src/index.tsx#L43-L50
  - https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/android/src/main/java/com/reactnativecommunityviewpager/ViewPagerViewManager.kt#L41
- Native swaps out blank pages with the newly supplied real pages
  - https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/android/src/main/java/com/reactnativecommunityviewpager/ViewPagerViewManager.kt#L57
  - https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/android/src/main/java/com/reactnativecommunityviewpager/ReactViewAdapter.kt#L47
  - https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/android/src/main/java/com/reactnativecommunityviewpager/ReactViewAdapter.kt#L12-L14
  - https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/android/src/main/java/com/reactnativecommunityviewpager/PageFragment.kt#L34-L44

troZee:
Thank you for an explanation. I will take a look, once I will be available. 

> For iOS, I noticed #140 . Should this lazy render implementation avoid UIPageViewController?

Lazy loading should also support iOS (UIPageViewController). For now, it is not on our roadmap to change iOS implementation of VP.

troZee:
I have tested your code and I would like to laso test more cases, which has been provided in our example. 

Your solution looks very interesting. 

When I was testing your code I used below example 

App.tsx
```
import React from 'react';
import { StyleSheet, Text, View } from 'react-native';
import { ViewPager } from '@react-native-community/viewpager';

const DummyComponent = ({ item }: { item: string }) => {
  React.useEffect(() => {
    console.log(`didmout ${item}`);
    return () => {
      console.log(`didunmout ${item}`);
    };
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
  return <Text style={styles.text}>{item}</Text>;
};

export default function App() {
  const data = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('');
  return (
    <ViewPager
      data={data}
      keyExtractor={(item) => item}
      renderItem={({ item }) => <DummyComponent item={item} />}
      style={styles.container}
    />
  );
}

const styles = StyleSheet.create({
  container: { flex: 1 },
  text: { fontSize: 100 },
});

```

and components are mounted and unmounted on page changed. Ofc I am wondering, if it is necessary to reuse pages, which has been created. 

JS side: 
- your solution looks good, but it breaks the API and probably it should be done on next branch instead of stable master version. 

Native side: 
- you don't need adapter inside the fragment https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/android/src/main/java/com/reactnativecommunityviewpager/ReactViewAdapter.kt#L31 . It creates memory leak 
- Please checkout `next` branch. We changed a way of passing views to the fragments
- Why observable ? https://github.com/alpha0010/react-native-viewpager/blob/09b188efa0162ea7335350fcd67ff63f4c73516a/android/src/main/java/com/reactnativecommunityviewpager/ReactViewAdapter.kt#L10


Summary: 
Lazy loading is handled on native side as I wrote above (android/ios). More for android here: https://www.programmersought.com/article/39134390787/

So you don't need to implement any solution to achieve lazy loading on native side. 

If you would like to continue your work, please follow below steps: 
- branch off of `next` branch
- apply changes 
- create new draft PR (PoC)

alpha0010:
Thanks for the review; I will take a look at what I can do.

PR #277, opened 2021-01-20
Lazy pager

### LINKED ISSUE #104: Dynamic / lazy approach to load pages

alpha0010:
# Summary

Enables pages to be rendered on demand, instead of all at once. Important when there are a very large number of pages.

Currently, a basic implementation. I just wiped the app, and init fresh with react-native-builder-bob, so will need to restore/rename some files later.

Continued from #276

## Test Plan

In progress...

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

Fixes #104


alpha0010:
@troZee I have tried to implement your Android suggestions.  I also drafted iOS-side.  Do you have any further feedback before I work on adding features (programmatic interaction, transition, etc.)?

troZee:
I have not got a chance to look at it. I will try to check in in next week.

android/src/main/java/com/reactnativecommunityviewpager/ReactViewAdapter.kt
```diff
@@ -0,0 +1,59 @@
...
+      }
+    }
+
+    // Let the ViewPager2 know which pages need to be updated. These are the
+    // pages that are now available from JS-side, and the pages that were just
+    // unmounted from JS-side.
+    // TODO: Currently assumes no pages will be inserted/deleted at an index
+    // before the currently displayed page.
+    for (position in changedPositions) {
+      notifyItemChanged(position)
```
troZee, line 48: question: you are using this logic to let react native know, something has changed ? 
Maybe it is better to memoize pages ?

android/src/main/java/com/reactnativecommunityviewpager/ViewPagerViewManager.kt
```diff
@@ -0,0 +1,70 @@
...
+
+  override fun createViewInstance(reactContext: ThemedReactContext): ViewPager2 {
+    return ViewPager2(reactContext).apply {
+      adapter = ReactViewAdapter(reactContext.currentActivity as FragmentActivity)
+
+      registerOnPageChangeCallback(object : OnPageChangeCallback() {
+        override fun onPageSelected(position: Int) {
+          reactContext.getJSModule(RCTEventEmitter::class.java).receiveEvent(
+            id,
+            "topPageSelected",
```
troZee, line 24: why `topPageSelected` ?

troZee:
Hey @alpha0010 ,

Thank you for creating a second PoC. Good job 🎉 

I have checked your draft and it looks much better. Android looks good to me and iOS has some glitches, which can be fixed (I can also help you to fix those issues on iOS). 

So now please adopt your changes to `next` branch and prepare PR, where I could see only changes. As you can see, this library uses Objc and Java and in case on any problem, feel free to ask. 

The second thing: I think, the components should be memoized bc for now, they are mounted and unmounted all the time, which is not very performant approach. 

Imagine, each page contains some api calls and your component is mounted and unmounted, when user changes the tabs (ofc you can use some stage management, but it is not a solution). In the same way works bottom tabs from react native, and it would be nice to have this kind of functionality. 

The second thing, is you don't need to notify native part in case on any change, bc this component is mounted only once. 

To sum it up, how I see it: 
- We have  `const data = 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'.split('');`
- And current it looks like below: 
```
didmout A
didmout B
didmout C
didunmout A
didmout D
didunmout B
didmout E
didunmout C
didmout F
didunmout D
didmout G
didunmout E
didmout H
didunmout F
didmout I
didunmout G
didmout J
didunmout H
didmout K
```
and expected is:
```
didmout A
didmout B
didmout C
didmout D
didmout E
didmout F
didmout G
didmout H
didmout I
didmout J
didmout K
```

for below component: 
```jsx
const DummyComponent = ({ item }: { item: string }) => {
  React.useEffect(() => {
    console.log(`didmout ${item}`);
    return () => {
      console.log(`didunmout ${item}`);
    };
    // eslint-disable-next-line react-hooks/exhaustive-deps
  }, []);
  return <Text style={styles.text}>{item}</Text>;
};

```

android/src/main/java/com/reactnativecommunityviewpager/ReactViewAdapter.kt
alpha0010: This is to say a new page (not previously rendered) now exists, or a page was removed (unmounted) js side.  Notifying newly rendered is important, since if the user switches pages quickly, the pager can end up displaying a blank page (since it is not yet rendered).  This lets the viewer reload the active (blank) page once the rendered content is available.

Notifying removes should help reduce memory consumption, if pages are very heavy.  (Although, I have not profiled, so am not certain this is the case.)

android/src/main/java/com/reactnativecommunityviewpager/ViewPagerViewManager.kt
alpha0010: If I understand correctly, that is what official docs state https://reactnative.dev/docs/native-components-android#events and how it is implemented on `next` https://github.com/callstack/react-native-viewpager/blob/next/android/src/main/java/com/reactnativecommunity/viewpager/event/PageSelectedEvent.java#L23 .

alpha0010:
> So now please adopt your changes to `next` branch and prepare PR, where I could see only changes. As you can see, this library uses Objc and Java and in case on any problem, feel free to ask.

To confirm I am understanding correctly, port this implementation to be directly edit over `next` branch?  Assuming so, I should be able to manage Java fine.  However, Objective-C is very hard for me (I can learn, but it will take some time).

> The second thing: I think, the components should be memoized bc for now, they are mounted and unmounted all the time, which is not very performant approach.
>
> Imagine, each page contains some api calls and your component is mounted and unmounted, when user changes the tabs (ofc you can use some stage management, but it is not a solution). In the same way works bottom tabs from react navigation, and it would be nice to have this kind of functionality.

Regarding avoiding unmounting, I worry about memory usage; my target use requires approximately 1000 pages.

This implementation currently renders a buffer of page before and after the currently visible page (buffer size is hard coded as 1 for simplicity of this PoC).  Does expanding the buffer, and reducing frequency of window shifts resolve the mount/unmount performance issue?
```
' ' - unmounted page
'.' - mounted page
'^' - currently visible page

ABCDEFGHIJKLMNOPQRSTUVWXYZ
^...

ABCDEFGHIJKLMNOPQRSTUVWXYZ
.^..

ABCDEFGHIJKLMNOPQRSTUVWXYZ
..^.

ABCDEFGHIJKLMNOPQRSTUVWXYZ
...^...

ABCDEFGHIJKLMNOPQRSTUVWXYZ
....^..

ABCDEFGHIJKLMNOPQRSTUVWXYZ
.....^.

ABCDEFGHIJKLMNOPQRSTUVWXYZ
   ...^...

ABCDEFGHIJKLMNOPQRSTUVWXYZ
   ....^..

ABCDEFGHIJKLMNOPQRSTUVWXYZ
   ...^...

ABCDEFGHIJKLMNOPQRSTUVWXYZ
   ..^....

ABCDEFGHIJKLMNOPQRSTUVWXYZ
   .^.....

ABCDEFGHIJKLMNOPQRSTUVWXYZ
...^...
```

troZee:
> To confirm I am understanding correctly, port this implementation to be directly edit over `next` branch? Assuming so, I should be able to manage Java fine.

Yes, please create a fresh branch without any changes from `next` and then make a changes only for Android. Then, I will try to help you convert swift into Objc. 


> Regarding avoiding unmounting, I worry about memory usage; my target use requires approximately 1000 pages.

I also was thinking about the ppl, who would like to avoid mounting/unmounting problem, so basically, we need to compromise it. 


>  Does expanding the buffer, and reducing frequency of window shifts resolve the mount/unmount performance issue?

No, bc if you increase buffer size, did mount is calling eagerly. I think we need to support two kinds of lazy loading in this case 
- with memoizing 
- without memoizing 

so your code should be modified a little bit like below (pseudocode): 
```jsx
    <ViewPagerNative
      count={props.data.length}
      offset={offset}
      onPageSelected={(event) => {
        setOffset(Math.max(0, event.nativeEvent.page - buffer));
      }}
      style={props.style}
    >
      {props.shouldMemoize ? memoizedData(props) : data(props)}
    </ViewPagerNative>
```

where `data` function would be your code:
```js
props.data.slice(offset, offset + windowLength).map((item, index) => (
            <View
              key={props.keyExtractor(item, offset + index)}
              style={styles.pageContainer}
            >
              {props.renderItem({ item, index: offset + index })}
            </View>
          ))
```

so in above case we are giving a ability to choose between memoization and creating new page each time.

PR #279, opened 2021-01-26
Lazy pager

### LINKED ISSUE #104: Dynamic / lazy approach to load pages

alpha0010:
# Summary

Enables pages to be rendered on demand, instead of all at once. Important when there are a very large number of pages.

Continued from #277

<details>

```diff
diff --git a/example/src/BasicPagerViewExample.tsx b/example/src/BasicPagerViewExample.tsx
index 99d49f0..7529e66 100644
--- a/example/src/BasicPagerViewExample.tsx
+++ b/example/src/BasicPagerViewExample.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { Image, SafeAreaView, StyleSheet, View } from 'react-native';
 
 import { PagerView } from 'react-native-pager-view';
@@ -8,7 +8,14 @@ import { NavigationPanel } from './component/NavigationPanel';
 import { useNavigationPanel } from './hook/useNavigationPanel';
 import { CreatePage, createPageKeyExtractor } from './utils';
 
-function renderItem({ item }: { item: CreatePage }) {
+function TrackingRender({ item }: { item: CreatePage }) {
+  useEffect(() => {
+    console.log(`didmout ${item.key}`);
+    return () => {
+      console.log(`didunmout ${item.key}`);
+    };
+  }, [item.key]);
+
   return (
     <View style={item.style}>
       <Image style={styles.image} source={item.imgSource} />
@@ -17,15 +24,21 @@ function renderItem({ item }: { item: CreatePage }) {
   );
 }
 
+function renderItem({ item }: { item: CreatePage }) {
+  return <TrackingRender item={item} />;
+}
+
 export function BasicPagerViewExample() {
-  const { ref, ...navigationPanel } = useNavigationPanel<CreatePage>();
+  const { ref, ...navigationPanel } = useNavigationPanel<CreatePage>(10000);
 
   return (
     <SafeAreaView style={styles.container}>
       <PagerView
+        buffer={2}
         ref={ref}
         style={styles.pagerView}
         scrollEnabled={navigationPanel.scrollEnabled}
+        maxRenderWindow={20}
         onPageScroll={navigationPanel.onPageScroll}
         onPageSelected={navigationPanel.onPageSelected}
         onPageScrollStateChanged={navigationPanel.onPageScrollStateChanged}
```

</details>

In the patched checkout, the basic example should now handle displaying 10000 pages (does not work in `master`).  Watching metro, page renders should show mounts/unmounts at reasonable times (e.g., swiping a few pages forward, then back again should not trigger any unmounts, but swiping many pages in a single direction (or jump to end) should unmount far away pages).

Also try (editing the basic example, letting updates push dynamically through metro) `orientation` and `transitionStyle` props.  These can be dynamically updated (dynamic updates for these props do not work in `master`).

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

Fixes #104


alpha0010:
@troZee Ported over the Kotlin implementation.  Should function equivalently to the PoC in #277.

Also updated the Basic Example to the new api (incomplete, as there are more features required to implement on the pager).

alpha0010:
When paging fast enough that the viewer attempts to display a page that is not yet rendered, a placeholder view is provided.  Once ready, the adapter calls `notifyItemChanged(index)` to trigger an update.  However, this does not seem to work on the actively viewed page (it does appear to work for updating non-visible pages).

I suppose rendering a larger buffer of pages could make it less likely to hit a blank, but that does not fix the underlying issue.

Any thoughts?

federicoeventifica:
This is great! I really need this. 

Do you have an estimate date to complete it and can I help you in any way? Thank you.

alpha0010:
Hard to estimate dates; I believe @troZee is working porting the Swift iOS implementation from #277 to Objective C, but uncertain of how much time he has available for this project.

I you have a handle on native Android code, help figuring out the `notifyItemChanged(index)` problem in my above comment would be greatly appreciated.  (Checkout this branch, `yarn bootstrap && yarn example android` to run the code.)

alpha0010:
I could not determine how to fix the Android render failure issue with react views managed directly by `ViewPager2`.  I reimplemented the wrapper from the first PoC (though hopefully without a memory leak this time).  I can revert if you think there is a better solution.

alpha0010:
Well, I have attempted Objective C.  Implementation not yet complete, but appears to be working.

alpha0010:
@troZee The features available in `next` that are not currently implemented here are:
- onPageScrollStateChanged
- keyboardDismissMode
- pageMargin
- offscreenPageLimit
- onMoveShouldSetResponderCapture
- showPageIndicator
- setScrollEnabled
- adding/removing pages (currently, changing the dataset may desync visible page, or even trigger a crash)

Are any/all of these blocking before this PR can be ready?  I ask because while this PR is unmerged, other PRs targeted at `next` will likely need to be rewritten (due to architecture changes here).

(There are still several bugs I am in the process of solving, so not ready until at least those are fixed.)

android/src/main/java/com/reactnativecommunity/viewpager/ViewPagerFragment.java
```diff
@@ -4,22 +4,58 @@
...
+        mPosition = position;
+        mReactViewRef = new WeakReference<>(child);
     }
 
-
     @Nullable
     @Override
     public View onCreateView(@NonNull LayoutInflater inflater, @Nullable ViewGroup container, @Nullable Bundle savedInstanceState) {
-        return view != null ? view : new View(getContext());
+        FrameLayout rootView = new FrameLayout(inflater.getContext());
```
troZee, line 30: Why react view is wrapped into framelayout ? 
Why we cannot directly return react view ?

troZee (changes requested):

troZee:
Hey @alpha0010 
FYI: `next` branch has been merged into master. You don't need to change a branch, bc next and master branch are identical. 

> Are any/all of these blocking before this PR can be ready? 

All of feature needs to be implemented. I will merge this PR, once it is finished. 

>  I ask because while this PR is unmerged, other PRs targeted at next will likely need to be rewritten (due to architecture changes here).

Due lots of braking changes, it will be targeted to the next branch as a experimental version. 
I would like to ask, if this api change is needed ? bc I don't see any reasons, why we should change the api.

android/src/main/java/com/reactnativecommunity/viewpager/ReactViewPagerManager.java
```diff
@@ -59,186 +46,121 @@ protected ViewPager2 createViewInstance(@NonNull ThemedReactContext reactContext
...
 
             @Override
             public void onPageSelected(int position) {
-                super.onPageSelected(position);
                 eventDispatcher.dispatchEvent(
                         new PageSelectedEvent(vp.getId(), position));
             }
-
-            @Override
-            public void onPageScrollStateChanged(int state) {
```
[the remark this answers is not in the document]
alpha0010: Was avoiding leaving stubs in my code, trying to add/fix one feature at a time. I had not yet returned to restore this event handler.

android/src/main/java/com/reactnativecommunity/viewpager/ReactViewPagerManager.java
```diff
@@ -59,186 +46,121 @@ protected ViewPager2 createViewInstance(@NonNull ThemedReactContext reactContext
...
+    public Map<String, Object> getExportedCustomDirectEventTypeConstants() {
+        return MapBuilder.<String, Object>of(
+                PageScrollEvent.EVENT_NAME, MapBuilder.of("registrationName", "onPageScroll"),
+                PageSelectedEvent.EVENT_NAME, MapBuilder.of("registrationName", "onPageSelected"));
     }
 
     @Override
-    public boolean needsCustomLayoutForChildren() {
-        return true;
+    protected void onAfterUpdateTransaction(@NonNull ViewPager2 view) {
```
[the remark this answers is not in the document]
alpha0010: RN calls this method after calling all applicable `@ReactProp` setters.  RN would not otherwise call `onAfterUpdateTransaction()` in the adapter, as it does not have knowledge of the adapter.

src/EagerPager.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+import { findNodeHandle, UIManager } from 'react-native';
+import type { EagerPagerProps } from './types';
+
+import { ViewPagerNative } from './ViewPagerNative';
+
+export class EagerPager extends React.PureComponent<EagerPagerProps> {
```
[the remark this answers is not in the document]
alpha0010: Mimics api of current version of viewpager; render all pages in place (as react children), as opposed to requiring managing a render function (similar to the ScrollView vs FlatList usage).  Component could technically be implemented user side, but is would be more efficient here, as it directly wraps native (instead of wrapping the complications of js-side lazy load).

android/src/main/java/com/reactnativecommunity/viewpager/ViewPagerFragment.java
alpha0010: Originally used the react view directly, however I switched in https://github.com/callstack/react-native-viewpager/pull/279/commits/610dcac0b2ab00cd0b4aa2b5225c5a347e32bd18 because I could not determine any other way to solve the blank page bug.  Android would get stuck on a white screen if the user paged fast enough for the screen not to be rendered before arriving.  I tried using `notifyItem*()` methods, but they did not help.

I would prefer using the react view directly, if there is some way to fix the blank screen bug.

android/src/main/java/com/reactnativecommunity/viewpager/ReactViewPagerManager.java
```diff
@@ -59,186 +46,121 @@ protected ViewPager2 createViewInstance(@NonNull ThemedReactContext reactContext
...
+        final FragmentAdapter adapter = (FragmentAdapter) parent.getAdapter();
+        adapter.addReactView(child, index);
+        parent.post(new Runnable() {
             @Override
             public void run() {
-                view.measure(
-                        View.MeasureSpec.makeMeasureSpec(view.getWidth(), View.MeasureSpec.EXACTLY),
-                        View.MeasureSpec.makeMeasureSpec(view.getHeight(), View.MeasureSpec.EXACTLY));
-                view.layout(view.getLeft(), view.getTop(), view.getRight(), view.getBottom());
+                adapter.onAfterUpdateTransaction();
```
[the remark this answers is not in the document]
alpha0010: I think that was due to it causing the screen to flash white, but I cannot recall for certain.

alpha0010:
> I would like to ask, if this api change is needed ? bc I don't see any reasons, why we should change the api.

User facing api should be nearly the same.  Requiring render prop is breaking.  Props controlling how many pages should be pre-rendered are new, but optional.  I would prefer to switch pager to named export, supplying both lazy and eager versions; this would be breaking, but is not required to do.
Large portions of pager internals are different due to requiring handling not-yet-rendered pages.

android/src/main/java/com/reactnativecommunity/viewpager/FragmentAdapter.java
```diff
@@ -1,72 +1,81 @@
...
+                : null;
     }
 
-    @Override
-    public boolean containsItem(long itemId) {
-        for(View child: childrenViews) {
-            if((int) itemId == child.getId()) {
-                return true;
+    public void onAfterUpdateTransaction() {
+        Iterator<WeakReference<ViewPagerFragment>> it = mFragments.iterator();
```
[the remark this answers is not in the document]
alpha0010: Goal is to swap newly available react views into the placeholder rendered fragments.  I could try to shift the logic to `onPageSelected`, but I worry that would result in scrolling to blank pages which would only appear once the user stops scrolling.

terrysahaidak:
I would really want to try this thing in https://github.com/terrysahaidak/react-native-gallery-toolkit

It's really hard to get the native pager animation right for both android and iOS. So I would want to just use the native implementation. But since it's a gallery thing, we might want to render thousands of pages at the same time and my current implementation with Reanimated works well. But the actual changing pages animation - doesn't.

alpha0010:
I am having trouble with excessive rerenders.  Applying the following, then in the basic example, the pager component sends a constant stream of `render()` log messages as I drag.  If I change to use the component directly (not wrapped as animated), this does not occur.  What is the reason for using the animated api here?  Am I missing something?

Strangely, even though the headphones example also uses animated, it does not trigger the excessive renders.

```diff
diff --git a/src/ViewPager.tsx b/src/ViewPager.tsx
index 705bec5..08d5484 100644
--- a/src/ViewPager.tsx
+++ b/src/ViewPager.tsx
@@ -173,6 +173,8 @@ export class ViewPager<ItemT> extends React.PureComponent<
   render() {
     const { offset, windowLength } = this.state;
 
+    console.log('render()');
+
     return (
       <ViewPagerNative
         count={this.props.data.length}
```

Any ideas?

troZee:
> I am having trouble with excessive rerenders. Applying the following, then in the basic example, the pager component sends a constant stream of `render()` log messages as I drag. If I change to use the component directly (not wrapped as animated), this does not occur. What is the reason for using the animated api here? Am I missing something?
> 
> Strangely, even though the headphones example also uses animated, it does not trigger the excessive renders.
> 
> ```diff
> diff --git a/src/ViewPager.tsx b/src/ViewPager.tsx
> index 705bec5..08d5484 100644
> --- a/src/ViewPager.tsx
> +++ b/src/ViewPager.tsx
> @@ -173,6 +173,8 @@ export class ViewPager<ItemT> extends React.PureComponent<
>    render() {
>      const { offset, windowLength } = this.state;
>  
> +    console.log('render()');
> +
>      return (
>        <ViewPagerNative
>          count={this.props.data.length}
> ```
> 
> Any ideas?

Are you able to reproduce it on viewpager component without animation ? Maybe additional renders come from animated props.

alpha0010:
> Are you able to reproduce it on viewpager component without animation ? Maybe additional renders come from animated props.

Reworked `useNavigationPanel()` not to use animated api, and changed the `onPageSelectedCallback` parameter to be optional (currently it is default initialized with `() => {}`, which means a new function generated every time). After this, `render()` calls reduced to a reasonable amount.

Unless you know of something gained by using animated api in the basic example, I think I will remove using animated there.

erkie:
Have you tried `why-did-you-render`? It's helped me spot a lot of missing `useCallback`, `useMemo` and other things:

```javascript
// wdyr.js
import React from 'react';

if (process.env.NODE_ENV === 'development') {
  const whyDidYouRender = require('@welldone-software/why-did-you-render');
  const ReactRedux = require('react-redux');

  whyDidYouRender(React, {
    trackAllPureComponents: true,
    trackExtraHooks: [
      [ReactRedux, 'useSelector', 'useMemo']
    ]
  });
}

// in index.js
import "./wdyr";

import { AppRegistry, Platform } from 'react-native';
import App from './App';
// ... res of app
```

alpha0010:
> Can you tell me, what left to implement ?

I believe the only remaining item is dynamic add/remove pages.  I am currently working on the Android implementation for that.

troZee:
> > Can you tell me, what left to implement ?
> 
> I believe the only remaining item is dynamic add/remove pages. I am currently working on the Android implementation for that.

Due misunderstanding and confusing, we needed to change a package name on npm. So for now, please implement all needed features and then we will try to update with master branch. I talked internally with the others developer and we decided to change android native part into kotlin. Sorry for confusion and for troubles from my side. 

This feature will land on `next` branch due lots of breaking changes as a alpha version. ViewPager has lots of dependent packages (https://www.npmjs.com/package/@react-native-community/viewpager) and I would like to avoid a big clash.  

I would like to release your changes as a alpha, so the other dev could check it easily in their internal projects (Some of issues are very specific and hard to reproduce) 

I would like to say big thank you @alpha0010 for your hard work and sorry once again for troubles

alpha0010:
> Due misunderstanding and confusing, we needed to change a package name on npm. So for now, please implement all needed features and then we will try to update with master branch. I talked internally with the others developer and we decided to change android native part into kotlin. Sorry for confusion and for troubles from my side.

I believe I have implemented all features.  Is there anything else you would like from my end currently?

troZee:
> > Due misunderstanding and confusing, we needed to change a package name on npm. So for now, please implement all needed features and then we will try to update with master branch. I talked internally with the others developer and we decided to change android native part into kotlin. Sorry for confusion and for troubles from my side.
> 
> I believe I have implemented all features.  Is there anything else you would like from my end currently?

Thank you for your hard work ❤️ could you update your branch with master and resolve the conflicts ?

alpha0010:
Merge complete.  (Kotlin code is mostly auto-converted, so may not be very "kotlin-y".)

src/types.ts
```diff
@@ -65,58 +76,68 @@ export interface PagerViewProps {
...
-  offscreenPageLimit?: number;
+  transitionStyle?: TransitionStyle;
+}
 
-  children: ReactNode;
+export interface PagerViewProps<ItemT> extends EagerPagerProps {
+  /**
+   * Number of pages to render before/after the current page. Minimum 1.
+   */
+  buffer?: number;
```
troZee, line 118: What is a difference between `offscreenPageLimit` and `buffer` ?

src/types.ts
```diff
@@ -22,18 +21,30 @@ export interface PageScrollStateChangedEvent {
   pageScrollState: PageScrollState;
 }
 
-export interface PagerViewProps {
+export interface EagerPagerProps {
```
troZee, line 24: I think, we should store all props in one interface. Could you merge `EagerPagerProps` and `PagerViewProps` together ?

src/types.ts
```diff
@@ -65,58 +76,68 @@ export interface PagerViewProps {
...
   /**
-   * Determines whether it's possible to overscroll a bit
-   * after reaching end or very beginning of pages.
+   * Render an item from `data` into a page.
    */
-  overdrag?: boolean;
+  renderItem: (info: { item: ItemT; index: number }) => React.ReactElement;
 }
+
+export type PagerViewState = { offset: number; windowLength: number };
```
troZee, line 143: Why this type is needed ?

src/utils.tsx
```diff
@@ -8,30 +8,17 @@
...
  */
 
-import React, { Children, ReactNode, ReactElement } from 'react';
+import { Children, ReactNode } from 'react';
 
-export const childrenWithOverriddenStyle = (children?: ReactNode) => {
-  // Override styles so that each page will fill the parent. Native component
-  // will handle positioning of elements, so it's not important to offset
-  // them correctly.
-  return Children.map(children, (child) => {
```
troZee, line 17: Why old implementation has been removed ?

src/PagerViewNative.tsx
```diff
@@ -1,17 +1,37 @@
...
+  overdrag?: boolean;
+  pageMargin?: number;
+  scrollEnabled?: boolean;
+  showPageIndicator?: boolean;
+  style: StyleProp<ViewStyle>;
+  transitionStyle?: TransitionStyle;
+};
 
-export const PagerViewViewManager = requireNativeComponent(
+export const PagerViewNative = requireNativeComponent<PagerViewNativeProps>(
```
troZee, line 35:
```suggestion
export const PagerViewViewManager = requireNativeComponent<PagerViewNativeProps>(
```

src/PagerViewNative.tsx
```diff
@@ -1,17 +1,37 @@
...
+  PageScrollStateChangedNativeEvent,
+  TransitionStyle,
+} from './types';
 
 const VIEW_MANAGER_NAME = 'RNCViewPager';
 
-interface PagerViewViewManagerType extends HostComponent<PagerViewProps> {
-  getInnerViewNode(): ReactElement;
-}
+type PagerViewNativeProps = {
```
troZee, line 17: Why old implementation has been removed ? we don't need to copy/paste the same types

src/PagerView.tsx
```diff
@@ -1,155 +1,207 @@
...
+        transitionStyle={this.props.transitionStyle}
+      >
+        {children}
+      </PagerViewNative>
     );
   }
 }
+
+const styles = StyleSheet.create({
+  pageContainer: { height: '100%', position: 'absolute', width: '100%' },
```
troZee, line 206: We don't need this style since lib has `childrenWithOverriddenStyle` method

src/PagerView.tsx
```diff
@@ -1,155 +1,207 @@
...
+   * A helper function to enable/disable scroll imperatively.
+   * The recommended way is using the scrollEnabled prop, however, there might
+   * be a case where an imperative solution is more useful (e.g. for not
+   * blocking an animation)
    */
-  public setScrollEnabled = (scrollEnabled: boolean) => {
+  setScrollEnabled(scrollEnabled: boolean) {
     UIManager.dispatchViewManagerCommand(
-      ReactNative.findNodeHandle(this),
-      getViewManagerConfig().Commands.setScrollEnabled,
```
troZee, line 132: Why `getViewManagerConfig().Commands.setScrollEnabled,` has been changed to `setScrollEnabled`

src/PagerView.tsx
```diff
@@ -1,155 +1,207 @@
...
+        currentPage: page,
+        maxRenderWindow: this.props.maxRenderWindow,
+        offset: prevState.offset,
+        windowLength: prevState.windowLength,
+      })
     );
-  };
+    // Send paging command.
+    requestAnimationFrame(() => {
+      UIManager.dispatchViewManagerCommand(findNodeHandle(this), 'setPage', [
```
troZee, line 75: why 'setPage' instead of `getViewManagerConfig().Commands.setPage,`

src/PagerView.tsx
```diff
@@ -1,155 +1,207 @@
...
+      offset: 0,
+      windowLength: 0,
+    });
+  }
 
-  private _onPageSelected = (e: PagerViewOnPageSelectedEvent) => {
-    if (this.props.onPageSelected) {
-      this.props.onPageSelected(e);
+  componentDidMount() {
+    if (this.props.initialPage != null && this.props.initialPage > 0) {
```
troZee, line 45:
```suggestion
    if (this.props.initialPage) {
```

troZee:
@alpha0010 
Could you provide me a test plan for your feature ?

android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
```diff
@@ -1,56 +1,152 @@
...
-    childrenViews.clear()
-    notifyDataSetChanged()
-  }
-
-  fun getChildViewAt(index: Int): View {
-    return childrenViews[index]
-  }
+    override fun createFragment(position: Int): Fragment {
+        val fragment = ViewPagerFragment(position, getViewAtPosition(position))
+        mFragments.add(WeakReference(fragment))
```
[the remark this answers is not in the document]
alpha0010: If the data set changes, required to notify fragments of what page index they are.  If a fragment is created before its content (the react view) is rendered, allows the fragment to update to display content when ready (I tried to use just `notifyItemChanged()` for these updates, but it did not work for me when the page being updated was the actively displayed page).

- https://github.com/alpha0010/react-native-viewpager/blob/a13b821be982b686e919fe1db836304cd3df6800/android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt#L123-L126
- https://github.com/alpha0010/react-native-viewpager/blob/a13b821be982b686e919fe1db836304cd3df6800/android/src/main/java/com/reactnativepagerview/ViewPagerFragment.kt#L33
- https://github.com/alpha0010/react-native-viewpager/blob/a13b821be982b686e919fe1db836304cd3df6800/android/src/main/java/com/reactnativepagerview/ViewPagerFragment.kt#L45

src/types.ts
alpha0010: `offscreenPageLimit` is an Android only concept, of how many (offscreen) fragments are held in native side memory.  iOS does not exactly have an equivalent.  The iOS native side pager (essentially) only ever holds at most three pages: previous (if not at beginning), current, and next (if not at end).

`buffer` controls how many views are rendered js side.  These react views are sent as a batch to the native side, which uses them to populate pages.  The smaller the `buffer`, the more likely a user could swipe to a page that does not yet have its content ready to display; native code will display a placeholder (blank page) until the next batch of rendered views arrives from js side.  This is the work around for native code not being able to synchronously request a specific page be rendered.

src/PagerView.tsx
alpha0010: Was trying to work around https://github.com/facebook/react-native/commit/6eec39313dc6a088bc870a47de3a74c69714d4c9 .  However, perhaps that is Facebook internal only.  I can revert.

src/types.ts
alpha0010: It is used to keep track, js side, of which views to render for native side.
- https://github.com/alpha0010/react-native-viewpager/blob/a13b821be982b686e919fe1db836304cd3df6800/src/PagerView.tsx#L147-L155
- https://github.com/alpha0010/react-native-viewpager/blob/a13b821be982b686e919fe1db836304cd3df6800/src/PagerView.tsx#L179

src/types.ts
alpha0010: The problem with that, is it would require `data`, `keyExtractor`, and `renderItem` be passed to the `EagerPager`, which does not use them.  It would also mean `buffer` and `maxRenderWindow` are accepted props, but are in reality ignored by `EagerPager`.

src/PagerViewNative.tsx
alpha0010: The native component and js component take different props.  Native component requires `childrenKeys`, `count`, `offset`, and `onMoveShouldSetResponderCapture`, which js component does not accept.  Js component accepts `initialPage`, `keyboardDismissMode`, `data`, `keyExtractor`, `maxRenderWindow`, and `renderItem`, which native component does not accept.

I could factor out the common props into a new type to inherit from, to avoid duplication.

src/PagerView.tsx
alpha0010: My thought was this gives more expected behavior to library users.

From the readme:
> **Attention:** Note that you can only use `View` components as children of `PagerView` (cf. folder */example*)
. For Android if `View` has own children, set prop `collapsable` to false <https://reactnative.dev/img/view#collapsable>, otherwise react-native might remove those children views and  and it's children will be rendered as separate pages

Avoiding using `childrenWithOverriddenStyle()` removes all of those limitations.

src/utils.tsx
alpha0010: See https://github.com/callstack/react-native-pager-view/pull/279#discussion_r596267728.

alpha0010:
> Could you provide me a test plan for your feature ?

Added in PR description.

hengkx:
Thank you very much for your work. I really need this feature.

mohshbool:
> Thank you very much for your work. I really need this feature.

How did you manage to implement it in the current version?

hengkx:
> > Thank you very much for your work. I really need this feature.
> 
> How did you manage to implement it in the current version?

https://github.com/callstack/react-native-pager-view/issues/104#issuecomment-679122828

mohshbool:
> > > Thank you very much for your work. I really need this feature.
> > 
> > 
> > How did you manage to implement it in the current version?
> 
> [#104 (comment)](https://github.com/callstack/react-native-pager-view/issues/104#issuecomment-679122828)

Thank you so much man!

Did you manage to solve this? `Property 'children' does not exist on type 'IntrinsicAttributes & LazyViewPagerProps<unknown> & RefAttributes<LazyViewPagerHandle>'.`

hengkx:
```
<LazyViewPager
          style={{ flex: 1, width }}
          initialPage={index}
          data={[]}
          onPageSelected={setPosition}
          renderItem={renderItem}
        />
```

mohshbool:
> ```
> <LazyViewPager
>           style={{ flex: 1, width }}
>           initialPage={index}
>           data={[]}
>           onPageSelected={setPosition}
>           renderItem={renderItem}
>         />
> ```

Oh, it's FlatList-like implementation.

That won't work for me, for the moment. Thank you though, you've been much help ❤️

mapss:
@alpha0010 @troZee  - thank you for your work. We are eagerly waiting for you To merge this PR. 🙏

alpha0010:
@troZee Have you had a chance to look at this?

android/src/main/java/com/reactnativepagerview/BiMap.kt
```diff
@@ -0,0 +1,26 @@
+package com.reactnativepagerview
+
+class BiMap<K, V> {
+    private val keyToValue = mutableMapOf<K, V>()
+    private val valueToKey = mutableMapOf<V, K>()
+
+    fun clear() {
```
nguyenminhbgbg, line 7: failed

android/src/main/java/com/reactnativepagerview/BiMap.kt
```diff
@@ -0,0 +1,26 @@
...
+        keyToValue.clear()
+        valueToKey.clear()
+    }
+
+    fun getByKey(key: K): V? {
```
nguyenminhbgbg, line 12: ok

nguyenminhbgbg (changes requested):
nêd change

troZee:
> When paging fast enough that the viewer attempts to display a page that is not yet rendered, a placeholder view is provided. Once ready, the adapter calls `notifyItemChanged(index)` to trigger an update. However, this does not seem to work on the actively viewed page (it does appear to work for updating non-visible pages).
> 
> I suppose rendering a larger buffer of pages could make it less likely to hit a blank, but that does not fix the underlying issue.
> 
> Any thoughts?

Below you can find an answer, why `notifyItemChanged` does not work:
https://stackoverflow.com/questions/44868899/recyclerview-in-react-native-notifyiteminserted-and-notifydatasetchanged-ha

https://stackoverflow.com/questions/49371866/recyclerview-wont-update-child-until-i-scroll

in our code, you can find it here: https://github.com/callstack/react-native-pager-view/blob/master/android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt#L64

android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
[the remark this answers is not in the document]
troZee: as I mentioned in the post `notifyItemChanged()` does not work in react native. Could you use below snippet to force update ? https://github.com/callstack/react-native-pager-view/blob/master/android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt#L64

troZee (changes requested):

troZee:
@alpha0010 
I would like to simplify implementation of Lazy View pager, bc for now, it has some dirty workarounds (like adding/removing view from frame layout). 

Recycler view is very specific and it does not see any changes, bc datasource is a view. Hence we can use 2 approach to fix view updating. 
1. https://stackoverflow.com/questions/49371866/recyclerview-wont-update-child-until-i-scroll
2. https://iammert.medium.com/using-diffutil-in-android-recyclerview-bdca8e4fbb00

Please try those two options and let me know the results

troZee (changes requested):

example/src/BasicPagerViewExample.tsx
```diff
@@ -1,46 +1,44 @@
...
-        style={styles.PagerView}
-        initialPage={0}
-        overdrag={navigationPanel.overdragEnabled}
+        style={styles.pagerView}
         scrollEnabled={navigationPanel.scrollEnabled}
         onPageScroll={navigationPanel.onPageScroll}
         onPageSelected={navigationPanel.onPageSelected}
         onPageScrollStateChanged={navigationPanel.onPageScrollStateChanged}
         pageMargin={10}
-        // Lib does not support dynamically orientation change
```
troZee, line 27: please revert it

example/src/BasicPagerViewExample.tsx
```diff
@@ -1,46 +1,44 @@
...
         orientation="horizontal"
-        // Lib does not support dynamically transitionStyle change
```
troZee, line 29: please revert it

troZee (changes requested):

troZee:
@alpha0010 
I tested the code for `maxRenderWindow={1}` and `buffer={2}` and I could see below values: 
```
[Thu Apr 08 2021 13:17:17.104]  LOG      didmout 0
[Thu Apr 08 2021 13:17:17.106]  LOG      didmout 1
[Thu Apr 08 2021 13:17:17.106]  LOG      didmout 2
[Thu Apr 08 2021 13:17:19.152]  LOG      didmout 3
[Thu Apr 08 2021 13:17:20.950]  LOG      didmout 4
[Thu Apr 08 2021 13:17:20.982]  LOG      didunmout 0
[Thu Apr 08 2021 13:17:21.120]  LOG      didmout 5
```

And it seems like `maxRenderWindow` does not work properly. Hence it can be confusing and  we should warn developer about the fact, that `maxRenderWindow` should be greater than `1 + 2 * buffer`, bc then it makes sense.

troZee:
@alpha0010 
I analyzed this PR and I did an experiment: 
I reverted whole android code and remain only JS part responsible for lazy loading. Below code in `src/PagerView.tsx`: 

```js
        .slice(offset, offset + windowLength)
```
breaks current android implementation. When I changed it to below code: 
```js
        .slice(0, offset + windowLength)
```

Everything works as expected without removing mounted views. 

Can we for now focus on JS part and implement lazy loading without removing mounted views ? 
Bc for now android implementation looks very convoluted and I would like to make it more simpler to understand

example/src/BasicPagerViewExample.tsx
alpha0010: In this branch it does work (both Android and iOS).

example/src/BasicPagerViewExample.tsx
alpha0010: In this branch it does work (iOS only; Android does not support any transition styles).

alpha0010:
Thanks for your review, I will get to work on those items.

> Can we for now focus on JS part and implement lazy loading without removing mounted views ?
> Bc for now android implementation looks very convoluted and I would like to make it more simpler to understand

I will create a separate PR for that.  To confirm, do you mean skip the Android native changes, and just use JS/iOS?  Or skip all native changes, and do only JS updates?

hengkx:
Will the native performance be better?

alpha0010:
> Will the native performance be better?

Using only JS changes, performance when near the beginning of a long list will be equatable to if also using native changes.  However, once farther into a long list of pages (say, maybe 50 pages in? depends on how heavy an individual page is), performance from only JS will start to drop (native would remain just as fast/low memory usage).

troZee:
> Thanks for your review, I will get to work on those items.
> 
> > Can we for now focus on JS part and implement lazy loading without removing mounted views ?
> > Bc for now android implementation looks very convoluted and I would like to make it more simpler to understand
> 
> I will create a separate PR for that. To confirm, do you mean skip the Android native changes, and just use JS/iOS? Or skip all native changes, and do only JS updates?

Please skip all native changes. I would be easier to me to divide PR into small chucks like JS/Android/iOS.

example/src/BasicPagerViewExample.tsx
troZee: Do you mean fast refresh ? bc if you pass a state here, it would not work.

example/src/BasicPagerViewExample.tsx
troZee: Do you mean fast refresh ? bc if you pass a state here, it would not work.

alpha0010: No, works on state change.  Can test by hacking the basic example; enable/disable dots will switch between horizontal/vertical, overdrag will switch between scroll/curl.

```diff
diff --git a/example/src/BasicPagerViewExample.tsx b/example/src/BasicPagerViewExample.tsx
index 99d49f0..295a78d 100644
--- a/example/src/BasicPagerViewExample.tsx
+++ b/example/src/BasicPagerViewExample.tsx
@@ -30,9 +30,9 @@ export function BasicPagerViewExample() {
         onPageSelected={navigationPanel.onPageSelected}
         onPageScrollStateChanged={navigationPanel.onPageScrollStateChanged}
         pageMargin={10}
-        orientation="horizontal"
+        orientation={navigationPanel.dotsEnabled ? 'horizontal' : 'vertical'}
         overdrag={navigationPanel.overdragEnabled}
-        transitionStyle="scroll"
+        transitionStyle={navigationPanel.overdragEnabled ? 'scroll' : 'curl'}
         showPageIndicator={navigationPanel.dotsEnabled}
         data={navigationPanel.pages}
         keyExtractor={createPageKeyExtractor}
```

erkie:
Just wanted to say that we implemented this in our app and it works fantastically. Thank you so much for your awesome work, it made a world of a difference. Specially for Android.  We'd love to buy you a coffee or donate something to your favourite charity as a thank you?

alpha0010:
Great to hear this has helped you.  I think the best thanks would be to "pay it forward" - fix a bug/add a feature in some open source library you use.  No matter how small a fix, it will make someone's day better, and keeps the community going.

Hannes-Endare:
How far are we from merging this? Would really love to use this feature :)

isidoro98:
Any updates on this? @alpha0010 @troZee

troZee:
> Any updates on this? @alpha0010 @troZee

This PR has been split into smaller chunks, and I will check all of them in my free time.

alpha0010:
> Any update on the PR, when it will be get merged ?

Last piece of this is in https://github.com/callstack/react-native-pager-view/pull/332 .  Hopefully I will have time soon to finish the remain requested changes.

troZee:
All PRs referring to Lazy VP has been merged ❤️  🎉 

Is it anything, that needs to be implemented ? @alpha0010 
 
Can we close this issue ? 

Below you can find a plan of release this feature: 
- check, if all stuffs has been implemented
- sync next with master branch (probably, I will need your help @alpha0010 )
- release 6.x-rcx on next tag 
- create a new thread, which refers to all issues from the next branch 
- if all issue has been resolved, then it will be merged to the main branch

alpha0010:
> Is it anything, that needs to be implemented ?

I believe everything is implemented

> Can we close this issue ?

Yes.

>     * sync next with master branch

https://github.com/callstack/react-native-pager-view/pull/371

troZee:
I am closing this issue bc everything has been implemented

troZee:
If you would like to test it inside your project, please use this version: 

Issue #283, opened 2021-02-08
[Android] Crash: NoSuchMethodException ViewPagerFragment.<init>


LenaZhuk:
## Environment

"react-native": "0.62.2"
"@react-native-community/viewpager": "^5.0.5"

Devices: Samsung Galaxy A31 (Android 10), Samsung Galaxy A11 (Android 10), Samsung Galaxy J5 (Android 9)

## Description

Crash in Firebase Crashlytics report:

Caused by java.lang.NoSuchMethodException
com.reactnativecommunity.viewpager.ViewPagerFragment.<init> []

## Stack trace

Caused by java.lang.NoSuchMethodException: com.reactnativecommunity.viewpager.ViewPagerFragment.<init> []
       at java.lang.Class.getConstructor0(Class.java:2332)
       at java.lang.Class.getConstructor(Class.java:1728)
       at androidx.fragment.app.Fragment.instantiate(Fragment.java:548)
       at androidx.fragment.app.FragmentContainer.instantiate(FragmentContainer.java:57)
       at androidx.fragment.app.FragmentManager$3.instantiate(FragmentManager.java:390)
       at androidx.fragment.app.FragmentStateManager.<init>(FragmentStateManager.java:74)
       at androidx.fragment.app.FragmentManager.restoreSaveState(FragmentManager.java:2454)
       at androidx.fragment.app.FragmentController.restoreSaveState(FragmentController.java:196)
       at androidx.fragment.app.FragmentActivity.onCreate(FragmentActivity.java:287)
       at androidx.appcompat.app.AppCompatActivity.onCreate(AppCompatActivity.java:115)
       at com.facebook.react.ReactActivity.onCreate(ReactActivity.java:44)
       at com.ellorem.oxygean.MainActivity.onCreate(MainActivity.java:15)
       at android.app.Activity.performCreate(Activity.java:7963)
       at android.app.Activity.performCreate(Activity.java:7952)
       at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1307)
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3629)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3806)
       at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:83)
       at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
[...]

Issue #282, opened 2021-02-08
[Android] Crash: could not find Fragment constructor


LenaZhuk:
## Environment

"react-native": "0.62.2"
"@react-native-community/viewpager": "^5.0.5"

Devices: Samsung Galaxy A31 (Android 10), Samsung Galaxy A11 (Android 10), Samsung Galaxy J5 (Android 9)

## Description

Crash in Firebase Crashlytics report:

Caused by androidx.fragment.app.Fragment$InstantiationException
Unable to instantiate fragment com.reactnativecommunity.viewpager.ViewPagerFragment: could not find Fragment constructor

## Stack trace

Caused by androidx.fragment.app.Fragment$InstantiationException: Unable to instantiate fragment com.reactnativecommunity.viewpager.ViewPagerFragment: could not find Fragment constructor
       at androidx.fragment.app.Fragment.instantiate(Fragment.java:563)
       at androidx.fragment.app.FragmentContainer.instantiate(FragmentContainer.java:57)
       at androidx.fragment.app.FragmentManager$3.instantiate(FragmentManager.java:390)
       at androidx.fragment.app.FragmentStateManager.<init>(FragmentStateManager.java:74)
       at androidx.fragment.app.FragmentManager.restoreSaveState(FragmentManager.java:2454)
       at androidx.fragment.app.FragmentController.restoreSaveState(FragmentController.java:196)
       at androidx.fragment.app.FragmentActivity.onCreate(FragmentActivity.java:287)
       at androidx.appcompat.app.AppCompatActivity.onCreate(AppCompatActivity.java:115)
       at com.facebook.react.ReactActivity.onCreate(ReactActivity.java:44)
       at com.ellorem.oxygean.MainActivity.onCreate(MainActivity.java:15)
       at android.app.Activity.performCreate(Activity.java:7963)
       at android.app.Activity.performCreate(Activity.java:7952)
       at android.app.Instrumentation.callActivityOnCreate(Instrumentation.java:1307)
       at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3629)
       at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3806)
       at android.app.servertransaction.LaunchActivityItem.execute(LaunchActivityItem.java:83)
       at android.app.servertransaction.TransactionExecutor.executeCallbacks(TransactionExecutor.java:135)
       at android.app.servertransaction.TransactionExecutor.execute(TransactionExecutor.java:95)
[...]

PR #285, opened 2021-02-09
chore: add empty constructor

arekkubaczkowski:
# Summary

Basically it's an attempt to fix an issue reported in #283 and #282 which are seems to be related with restoring fragments state and creating ViewPagerFragment class without arguments.
Unfortunately, because this error not occurs on local environment, before we'll try to find some better solution we want to make sure that our guess is proper by implementing constructor without any arguments and check if it's helped.


troZee:
@LenaZhuk 

Could you test it, before I merge it ?

troZee (approved):

LenaZhuk:
@troZee yes, we can try to test our app for a few days using the ViewPager from this branch to see if Firebase still logs the crashes. But most of the crashes came from our users, so this testing may not be accurate.

PR #295, opened 2021-03-02
fix (android): crash when scroll during unmount

ku8ar:
# Summary

Similar bug to this: https://github.com/callstack/react-native-viewpager/pull/271

Longcat: `Scrapped or attached views may not be recycled`

Fix just removes touch event execution when component is unmounting. Maybe there is some nicer way to fix this error, but I haven't found it...

## Test Plan

Render the instance of ViewPager inside of createNativeStackNavigator screen.
Enter to the screen and try to leave it by clicking back and then (very fast) try to swipe (during the screen removal animation).

It's hard to reproduce this bug in simulator. The easiest way is to just add timeout to the screen with `goBack()` running after a few seconds.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅    |

## Checklist


troZee:
https://github.com/callstack/react-native-viewpager/releases/tag/v5.0.10

PR #301, opened 2021-03-08
feat: support negative values for pageMargin on Android

satya164:
Supporting negative values let's us have pages peek from the previous and next pages.


Norcy:
Can iOS support negative margin?

troZee:
> Can iOS support negative margin?

Unfortunately we cannot support negative margins for iOS, bc iOS component creates each screen in lazy way, hence there is no option to render the previous and the next page. More you can find here: https://github.com/callstack/react-native-pager-view/issues/140

PR #305, opened 2021-03-09
chore: change package name

troZee:
# Summary

Move library out of @react-native-community scope. We needed to change name to `react-native-pager-view`, bc `eact-native-viewpager` was not free on the npm. I renamed all stuffs from ViewPager to PagerView to keep constistency and avoid any confusion. I also migrated the native part from java to kotlin to keep library up to date and there is no braking changes with these change. I was not sure, how about the swift migration. That matter is under consideration. For now native stack will be kotlin/objc. 

https://github.com/react-native-community/discussions-and-proposals/issues/330

## Test Plan

Just run an example

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅   |

## Checklist

## TODO

- [x] Add migration section 
- [x] Add information about previous package name


chirag04:
maybe keep the name but publish under callstack? @callstack/react-native-viewpager

troZee:
> maybe keep the name but publish under callstack? @callstack/react-native-viewpager

This library comes from Lean Core https://github.com/facebook/react-native/issues/23313 , so we would like to publish it outside callstack package. We only publish our tools and internal packages under @callstack scope

PR #331, opened 2021-04-14
Lazy pager (Android native side)

alpha0010:
# Summary

Enables pages to be rendered on demand, instead of all at once. This is the Android native code changes.

Split from #279. Note that this builds on top of #326, so when reviewing code, may be easier to check the specific commits that are not in #326.

<details>

```diff
diff --git a/example/src/BasicPagerViewExample.tsx b/example/src/BasicPagerViewExample.tsx
index 48ad5a4..69df7a3 100644
--- a/example/src/BasicPagerViewExample.tsx
+++ b/example/src/BasicPagerViewExample.tsx
@@ -15,6 +15,17 @@ function keyExtractor(page: CreatePage) {
 }
 
 function renderItem({ item }: { item: CreatePage }) {
+  return <TrackingRender item={item} />;
+}
+
+function TrackingRender({ item }: { item: CreatePage }) {
+  React.useEffect(() => {
+    console.log(`didmout ${item.key}`);
+    return () => {
+      console.log(`didunmout ${item.key}`);
+    };
+  }, [item.key]);
+
   return (
     <View style={item.style}>
       <Image style={styles.image} source={item.imgSource} />
@@ -26,16 +37,18 @@ function renderItem({ item }: { item: CreatePage }) {
 export function BasicPagerViewExample() {
   const { ref, ...navigationPanel } = useNavigationPanel<
     LazyPagerView<unknown>
-  >();
+  >(10000);
 
   return (
     <SafeAreaView style={styles.container}>
       <AnimatedPagerView
+        buffer={2}
         ref={ref}
         style={styles.PagerView}
         initialPage={0}
         overdrag={navigationPanel.overdragEnabled}
         scrollEnabled={navigationPanel.scrollEnabled}
+        maxRenderWindow={20}
         onPageScroll={navigationPanel.onPageScroll}
         onPageSelected={navigationPanel.onPageSelected}
         onPageScrollStateChanged={navigationPanel.onPageScrollStateChanged}
```

</details>

In the patched checkout, the basic example should now handle displaying 10000 pages (does not work in `master`). Watching metro, page renders should show mounts/unmounts at reasonable times (e.g., swiping a few pages forward, then back again should not trigger any unmounts, but swiping many pages in a single direction (or jump to end) should unmount far away pages).

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
+++ b/android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
@@ -3,54 +3,98 @@ package com.reactnativepagerview
 import android.view.View
 import androidx.fragment.app.Fragment
 import androidx.fragment.app.FragmentActivity
+import androidx.recyclerview.widget.DiffUtil
 import androidx.viewpager2.adapter.FragmentStateAdapter
-import java.util.*
 
 
 class FragmentAdapter(fragmentActivity: FragmentActivity) : FragmentStateAdapter(fragmentActivity) {
-  private val childrenViews: MutableList<View> = ArrayList()
+  private val childrenViews = mutableListOf<View>()
+  private var count = 0
+  private var offset = 0
+  private var isDirty = false
+  private val prevItemIds = mutableListOf<Long>()
+
   override fun createFragment(position: Int): Fragment {
-    return ViewPagerFragment(childrenViews[position])
+    return ViewPagerFragment(getViewAtPosition(position))
   }
 
-  override fun getItemCount(): Int {
-    return childrenViews.size
-  }
+  override fun getItemCount() = count
 
   override fun getItemId(position: Int): Long {
-    return childrenViews[position].id.toLong()
+    val view = getViewAtPosition(position)
+    return view?.id?.toLong() ?: UNRENDERED_ID_OFFSET + position
   }
 
   override fun containsItem(itemId: Long): Boolean {
-    for (child in childrenViews) {
-      if (itemId.toInt() == child.id) {
-        return true
-      }
+    if (itemId >= UNRENDERED_ID_OFFSET) {
+      val position = itemId - UNRENDERED_ID_OFFSET
+      return getViewAtPosition(position.toInt()) == null
     }
-    return false
+    return childrenViews.any { it.id.toLong() == itemId }
   }
 
-  fun addFragment(child: View, index: Int) {
+  /**
+   * Returns true if any changes were applied.
+   */
+  fun applyChanges(): Boolean {
+    if (!isDirty) {
+      return false
+    }
+
+    isDirty = false
+    val diff = DiffUtil.calculateDiff(
+      PagerDiffCallback(prevItemIds, this),
+      false
+    )
+    diff.dispatchUpdatesTo(this)
+    return true
+  }
+
+  fun setCount(count: Int) {
+    if (this.count != count) {
+      markDirty()
+      this.count = count
+    }
+  }
+
+  fun setOffset(offset: Int) {
+    if (this.offset != offset) {
+      markDirty()
+      this.offset = offset
+    }
+  }
+
+  fun addReactView(child: View, index: Int) {
+    markDirty()
     childrenViews.add(index, child)
-    notifyItemInserted(index)
   }
 
-  fun removeFragment(child: View) {
-    val index = childrenViews.indexOf(child)
-    removeFragmentAt(index)
-  }
+  fun getReactChildAt(index: Int) = childrenViews[index]
 
-  fun removeFragmentAt(index: Int) {
+  fun getReactChildCount() = childrenViews.size
+
+  fun removeReactViewAt(index: Int) {
+    markDirty()
     childrenViews.removeAt(index)
-    notifyItemRemoved(index)
   }
 
-  fun removeAll() {
-    childrenViews.clear()
-    notifyDataSetChanged()
+  private fun getViewAtPosition(position: Int): View? {
+    val index = position - offset
+    return if (index >= 0 && index < childrenViews.size) childrenViews[index] else null
   }
 
-  fun getChildViewAt(index: Int): View {
-    return childrenViews[index]
+  private fun markDirty() {
+    if (isDirty) {
+      return
+    }
+    isDirty = true
+    prevItemIds.clear()
+    for (position in 0 until itemCount) {
+      prevItemIds.add(getItemId(position))
+    }
+  }
+
+  companion object {
+     const val UNRENDERED_ID_OFFSET = 0xffffffffL
   }
 }
--- a/android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
+++ b/android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
@@ -36,7 +36,7 @@ class FragmentAdapter(fragmentActivity: FragmentActivity) : FragmentStateAdapter
   /**
    * Returns true if any changes were applied.
    */
-  fun applyChanges(): Boolean {
+  fun notifyAboutChanges(): Boolean {
     if (!isDirty) {
       return false
     }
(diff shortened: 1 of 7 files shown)


android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
```diff
@@ -3,54 +3,98 @@ package com.reactnativepagerview
...
+    prevItemIds.clear()
+    for (position in 0 until itemCount) {
+      prevItemIds.add(getItemId(position))
+    }
   }
 
-  fun getChildViewAt(index: Int): View {
-    return childrenViews[index]
+  companion object {
+     const val UNRENDERED_ID_OFFSET = 0xffffffffL
```
troZee, line 98: Could you explain me the purpose of this static val ?

android/src/main/java/com/reactnativepagerview/PagerDiffCallback.kt
```diff
@@ -0,0 +1,24 @@
...
+  override fun getNewListSize() = adapter.itemCount
+
+  override fun areItemsTheSame(oldItemPosition: Int, newItemPosition: Int): Boolean {
+    val oldId = oldList[oldItemPosition]
+    val newId = adapter.getItemId(newItemPosition)
+    if (oldId >= UNRENDERED_ID_OFFSET || newId >= UNRENDERED_ID_OFFSET) {
+      // An unrendered item is assumed the same as any item in the same position.
+      return oldItemPosition == newItemPosition
+    }
+    return oldId == newId
```
troZee, line 18:
```suggestion
    return oldId >= UNRENDERED_ID_OFFSET || newId >= UNRENDERED_ID_OFFSET ? oldItemPosition == newItemPosition : oldId == newId
```

android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
```diff
@@ -3,54 +3,98 @@ package com.reactnativepagerview
...
+    val view = getViewAtPosition(position)
+    return view?.id?.toLong() ?: UNRENDERED_ID_OFFSET + position
   }
 
   override fun containsItem(itemId: Long): Boolean {
-    for (child in childrenViews) {
-      if (itemId.toInt() == child.id) {
-        return true
-      }
+    if (itemId >= UNRENDERED_ID_OFFSET) {
```
troZee, line 29: Could you explain me this logic ? bc why itemId needs to be greater than `UNRENDERED_ID_OFFSET`

android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
```diff
@@ -191,6 +198,16 @@ class PagerViewViewManager : ViewGroupManager<ViewPager2>() {
     }
   }
 
+  /**
+   * Helper to trigger ViewPager2 to update.
+   */
+  private fun reLayoutView(view: View) {
```
troZee, line 204:
```suggestion
  private fun updateLayoutView(view: View) {
```

android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
```diff
@@ -3,54 +3,98 @@ package com.reactnativepagerview
...
+    return childrenViews.any { it.id.toLong() == itemId }
   }
 
-  fun addFragment(child: View, index: Int) {
-    childrenViews.add(index, child)
-    notifyItemInserted(index)
+  /**
+   * Returns true if any changes were applied.
+   */
+  fun applyChanges(): Boolean {
```
troZee, line 39:
```suggestion
  fun notifyAboutChanges(): Boolean {
```

android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
```diff
@@ -61,49 +58,47 @@ class PagerViewViewManager : ViewGroupManager<ViewPager2>() {
...
+    view.post { reLayoutView(view) }
     view.setCurrentItem(selectedTab, scrollSmooth)
   }
 
   override fun addView(parent: ViewPager2, child: View, index: Int) {
-    if (child == null) {
-      return
+    val adapter = parent.adapter as FragmentAdapter
+    adapter.addReactView(child, index)
+    parent.post {
```
troZee, line 68: Can we move this code to separate function called `postNewChanges` ? I spotted it is repeated few times.

troZee (changes requested):
Hey, 
I also added few comments here. It looks some much better with Diff tools implementation. 
Before you fix this comments. Please focus first on JS part, then I will merge the JS part and after that, please sync your branch with next.

troZee:
@alpha0010 good job 🎉  ❤️

troZee:
@alpha0010 Could you explain me why we need android changes, when it is already implemented on JS side ?
Bc I understand it in below way: 
JS part slice the array of children based on window length, but android is responsible for what ?

alpha0010:
> @alpha0010 Could you explain me why we need android changes, when it is already implemented on JS side ?
> Bc I understand it in below way:
> JS part slice the array of children based on window length, but android is responsible for what ?

Android native changes enable (off screen) pages to unload without triggering a change in page count/page indexes, and enable navigating to a page before it has rendered.

With JS only:
- Lazy render must, at minimum, keep the current page and *every* page before the current page loaded. This makes it impossible to have more pages than can be kept rendered in memory at the same time (app would crash if scrolling to the last page), and means there is no performance improvement if using a high initial page.
- Swiping quickly can cause the pager to feel stuck when arriving at the end of the buffer, until next pages finish rendering.

These two issues can be [worked around](https://github.com/callstack/react-native-pager-view/issues/104#issuecomment-679122828) with some trickery, but that technique is imperfect.

android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
alpha0010: Every page in `ViewPager2` must have an id. The id must stably refer to the same page, otherwise data mutations could cause the pager to lose track of what is the current page.

The simplest way is to use the page's `View` id (this is how the current master branch works). RN manages the id stability.

However with lazy rendering, a specific page may not have a `View` available when asked to provide an id.  So instead (when the `View` is not available), compute the id as `UNRENDERED_ID_OFFSET + pageIndex`.  This guarantees a unique id since all `View` ids are less than `UNRENDERED_ID_OFFSET`.  The custom `PagerDiffCallback` uses this constant as a threshold to determine if a page is rendered, so when a new render arrives, the pager can swap the unrendered for rendered without losing track of what the current page is.

troZee:
> > @alpha0010 Could you explain me why we need android changes, when it is already implemented on JS side ?
> > Bc I understand it in below way:
> > JS part slice the array of children based on window length, but android is responsible for what ?
> 
> Android native changes enable (off screen) pages to unload without triggering a change in page count/page indexes, and enable navigating to a page before it has rendered.
> 
> With JS only:
> 
> * Lazy render must, at minimum, keep the current page and _every_ page before the current page loaded. This makes it impossible to have more pages than can be kept rendered in memory at the same time (app would crash if scrolling to the last page), and means there is no performance improvement if using a high initial page.
> * Swiping quickly can cause the pager to feel stuck when arriving at the end of the buffer, until next pages finish rendering.
> 
> These two issues can be [worked around](https://github.com/callstack/react-native-pager-view/issues/104#issuecomment-679122828) with some trickery, but that technique is imperfect.

I see. Thank you for an explanation.

android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
troZee: Thank you for an explanation :D

troZee:
@alpha0010 I merged the js part. Could you please resolve the conflicts ?

alpha0010:
Thanks.

I believe I have made your requested changes.

troZee:

PR #332, opened 2021-04-14
Lazy pager (iOS native side)

alpha0010:
# Summary

Enables pages to be rendered on demand, instead of all at once. This is the iOS native code changes.

Split from #279. Note that this builds on top of #331, so when reviewing code, may be easier to check the specific commits that are not in #331.

<details>

```diff
diff --git a/example/src/BasicPagerViewExample.tsx b/example/src/BasicPagerViewExample.tsx
index 48ad5a4..69df7a3 100644
--- a/example/src/BasicPagerViewExample.tsx
+++ b/example/src/BasicPagerViewExample.tsx
@@ -15,6 +15,17 @@ function keyExtractor(page: CreatePage) {
 }
 
 function renderItem({ item }: { item: CreatePage }) {
+  return <TrackingRender item={item} />;
+}
+
+function TrackingRender({ item }: { item: CreatePage }) {
+  React.useEffect(() => {
+    console.log(`didmout ${item.key}`);
+    return () => {
+      console.log(`didunmout ${item.key}`);
+    };
+  }, [item.key]);
+
   return (
     <View style={item.style}>
       <Image style={styles.image} source={item.imgSource} />
@@ -26,16 +37,18 @@ function renderItem({ item }: { item: CreatePage }) {
 export function BasicPagerViewExample() {
   const { ref, ...navigationPanel } = useNavigationPanel<
     LazyPagerView<unknown>
-  >();
+  >(10000);
 
   return (
     <SafeAreaView style={styles.container}>
       <AnimatedPagerView
+        buffer={2}
         ref={ref}
         style={styles.PagerView}
         initialPage={0}
         overdrag={navigationPanel.overdragEnabled}
         scrollEnabled={navigationPanel.scrollEnabled}
+        maxRenderWindow={20}
         onPageScroll={navigationPanel.onPageScroll}
         onPageSelected={navigationPanel.onPageSelected}
         onPageScrollStateChanged={navigationPanel.onPageScrollStateChanged}
```

</details>

In the patched checkout, the basic example should now handle displaying 10000 pages (does not work in `master`). Watching metro, page renders should show mounts/unmounts at reasonable times (e.g., swiping a few pages forward, then back again should not trigger any unmounts, but swiping many pages in a single direction (or jump to end) should unmount far away pages).

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -1,4 +1,3 @@
-
 #import "ReactNativePageView.h"
 #import "React/RCTLog.h"
 #import <React/RCTViewManager.h>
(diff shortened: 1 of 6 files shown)


troZee:
@alpha0010 could you update this PR ?

ios/ReactViewPagerManager.m
```diff
@@ -34,54 +32,30 @@ - (void) goToPage
     }];
 }
 
-- (void) changeScrollEnabled
```
troZee, line 37: why it has been removed ?

ios/ReactViewPagerManager.m
```diff
@@ -34,54 +32,30 @@ - (void) goToPage
...
-                  : (nonnull NSNumber *)enabled) {
-    BOOL isEnabled = [enabled boolValue];
-    [self changeScrollEnabled:reactTag enabled:isEnabled];
-}
-
-RCT_CUSTOM_VIEW_PROPERTY(scrollEnabled, BOOL, ReactNativePageView) {
-    [view shouldScroll:[RCTConvert BOOL:json]];
-}
-
-RCT_CUSTOM_VIEW_PROPERTY(keyboardDismissMode, NSString, ReactNativePageView) {
```
troZee, line 75: Why it has been removed ?

ios/ReactViewPagerManager.m
```diff
@@ -34,54 +32,30 @@ - (void) goToPage
...
-    [view shouldDismissKeyboard:[RCTConvert NSString:json]];
-}
-
-RCT_CUSTOM_VIEW_PROPERTY(showPageIndicator, BOOL, ReactNativePageView) {
```
troZee, line 79: Why it has been removed ?

ios/ReactViewPagerManager.m
```diff
@@ -34,54 +32,30 @@ - (void) goToPage
...
+            RCTLogError(@"Cannot find ReactNativePageView with tag #%@", reactTag);
+            return;
+        }
+        [view shouldScroll:scrollEnabled];
+    }];
 }
 
 - (UIView *)view {
-    return [[ReactNativePageView alloc] initWithEventDispatcher:self.bridge.eventDispatcher];
+    return [[ReactNativePageView alloc] init];
```
troZee, line 58: Why view is not initialized using a bridge ?

ios/ReactNativePageView.m
```diff
@@ -10,412 +9,333 @@
...
-    
-    self.lastContentOffset = scrollView.contentOffset;
-    [self.eventDispatcher sendEvent:[[RCTOnPageScrollEvent alloc] initWithReactTag:self.reactTag position:@(position) offset:@(absoluteOffset)]];
+    self.onPageScroll(@{
+        @"offset": [NSNumber numberWithFloat:offset],
+        @"position": [NSNumber numberWithInteger:position]
+    });
 }
 
-- (NSString *)determineScrollDirection:(UIScrollView *)scrollView {
```
troZee, line 404: Why it has been removed ?

ios/ReactNativePageView.m
```diff
@@ -10,412 +9,333 @@
...
-        //Below line fix bug, where the view does not update after orientation changed.
-        [self updateDataSource];
-    }
-}
-
-- (void)didUpdateReactSubviews {
-    if (!self.reactPageViewController && self.reactViewController != nil) {
+- (instancetype)init {
+    if (self = [super init]) {
+        self.controllerCache = [NSMapTable weakToWeakObjectsMapTable];
```
troZee, line 27: Why self instead of _ ?

ios/ReactNativePageView.m
```diff
@@ -10,412 +9,333 @@
...
-- (void)shouldDismissKeyboard:(NSString *)dismissKeyboard;
-
 
 @end
 
-@implementation ReactNativePageView {
-    uint16_t _coalescingKey;
-}
 
-- (instancetype)initWithEventDispatcher:(RCTEventDispatcher *)eventDispatcher {
```
troZee, line 35: Why it has been removed ?

ios/ReactNativePageView.m
```diff
@@ -10,412 +9,333 @@
...
     }
+    [self goTo:targetPage animated:NO];
 }
 
-- (void)didMoveToWindow {
-    [super didMoveToWindow];
-    if (!self.reactPageViewController && self.reactViewController != nil) {
+- (void)didSetProps:(NSArray<NSString *> *)changedProps {
+    if (
+        [changedProps containsObject:@"orientation"]
```
troZee, line 55: why additional logic regarding to props setting is added ?

ios/ReactNativePageView.m
```diff
@@ -10,412 +9,333 @@
...
 
-- (void)shouldScroll:(BOOL)scrollEnabled {
-    _scrollEnabled = scrollEnabled;
-    if (self.reactPageViewController.view) {
-        self.scrollView.scrollEnabled = scrollEnabled;
-    }
+    [self attachPageIndicator];
 }
 
-- (void)shouldDismissKeyboard:(NSString *)dismissKeyboard {
```
troZee, line 139: why it has been removed ?

ios/ReactNativePageView.m
```diff
@@ -10,412 +9,333 @@
...
-    CGPoint point = scrollView.contentOffset;
-
-    float offset = 0;
-    
-    if (!_overdrag) {
-        if (_currentIndex == 0 && scrollView.contentOffset.x < scrollView.bounds.size.width) {
-            scrollView.contentOffset = CGPointMake(scrollView.bounds.size.width, 0);
-        } else if (_currentIndex == _reactPageIndicatorView.numberOfPages - 1 && scrollView.contentOffset.x > scrollView.bounds.size.width) {
-            scrollView.contentOffset = CGPointMake(scrollView.bounds.size.width, 0);
+    if (!self.overdrag) {
```
troZee, line 297: Could you explain, why you added logic under if statement ?

troZee (changes requested):
Thank you so much for iOS implementation 🎉  I have added few comments to your PR

ios/ReactViewPagerManager.m
```diff
@@ -7,24 +6,23 @@ @implementation ReactViewPagerManager
 
 RCT_EXPORT_MODULE(RNCViewPager)
 
-RCT_EXPORT_VIEW_PROPERTY(initialPage, NSInteger)
```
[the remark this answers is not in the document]
alpha0010: For parity of implementation with Android; implement in js for both.

I suppose it might make more sense to instead implement on the native side for both.

ios/ReactNativePageView.m
alpha0010: `_` and `self` seemed to be used interchangeably, without any obvious reason for one over the other.  I choose to use just one, for consistency.

Looked it up now, and appears `_` is more efficient for basic assignment, since it bypasses setters.  I can switch to that.  However, I am quite unfamiliar with Objective C.  Are there places other than the constructor that should also be switched?

ios/ReactNativePageView.m
alpha0010: For switching rendered pages, the pager must do more maintenance on caches.  However, native props can change much more frequently in lazy mode, so prefer to do as little maintenance as possible.

This has the extra benefit of making it possible to dynamically switch orientation/transition without requiring unmounting the component.

ios/ReactNativePageView.m
alpha0010: Seemed unnecessary to me, since js-side already implemented.  I can add it back.

ios/ReactNativePageView.m
alpha0010: The function is not used by any other code.

ios/ReactViewPagerManager.m
alpha0010: Was handled in js.  I can convert it back to native.

ios/ReactViewPagerManager.m
alpha0010: It is moved to line 15, and changed to `RCT_EXPORT_VIEW_PROPERTY`, since custom was unnecessary.

ios/ReactViewPagerManager.m
alpha0010: It was a while ago when I wrote this, but I think it was because the bridge was not used.

ios/ReactViewPagerManager.m
alpha0010: Defined it inline, as opposed to a separate function.  I think it might have occurred during my Swift -> Objective C conversion?

I can extract it again, if you prefer.

ios/ReactNativePageView.m
troZee: I did not notice is it implemented on the js part, hence it is not needed

ios/ReactNativePageView.m
troZee: Thank you for an explanation

ios/ReactNativePageView.m
troZee: > `_` and `self` seemed to be used interchangeably, without any obvious reason for one over the other. I choose to use just one, for consistency.
> 
> Looked it up now, and appears `_` is more efficient for basic assignment, since it bypasses setters. I can switch to that. However, I am quite unfamiliar with Objective C. Are there places other than the constructor that should also be switched?

I am not familiar with objC too, but I would prefer to use `_` in each place instead of `self`. Could you revert it to _ ? 

> Are there places other than the constructor that should also be switched?

I think, the whole file should contain `_` to keep consistency

ios/ReactViewPagerManager.m
troZee: The bridge is used to send the gesture events to the native side

ios/ReactViewPagerManager.m
troZee: No needed :D

ios/ReactViewPagerManager.m
troZee: It can be defined inline

troZee:
@alpha0010 
I noticed, that frame is not correct measured on the first render 
Simulator Screen Shot - iPhone 11 - 2021-05-24 at 10 26 20

When I swipe to the right, everything is fine 
Simulator Screen Shot - iPhone 11 - 2021-05-24 at 10 26 24

I resolved this issue calling this method in the layoutSubviews 
https://github.com/callstack/react-native-pager-view/blob/master/ios/ReactNativePageView.m#L56

troZee (changes requested):
Thank you for your hard work, I found some issue, which has been described above

alpha0010:
A bit busy with other projects at the moment, but I should get back soon.

troZee (approved):

troZee:

PR #363, opened 2021-05-27
fix(ios): offset is changing  while overDrag is disabled

DanijelBojcic:
# Summary

When disabling overDrag on iOs and swiping right on the first page to trigger it the `offset` would change between `0` and `0.05...`. This can be seen on the Headphone example. ( need to add `overDrag={false}` prop to the pager component ) 

I fixed it by setting the `offset` to `0` when it goes into the `!_overDrag` logic.

## Test Plan
Before:

https://user-images.githubusercontent.com/50402950/119804695-e17dc580-bee0-11eb-9946-d080ba9c945e.mp4

After:

https://user-images.githubusercontent.com/50402950/119805009-299ce800-bee1-11eb-968e-1cecf723b710.mp4

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -363,14 +363,6 @@
 
     float offset = 0;
     
-    if (!_overdrag) {
-        if (_currentIndex == 0 && scrollView.contentOffset.x < scrollView.bounds.size.width) {
-            scrollView.contentOffset = CGPointMake(scrollView.bounds.size.width, 0);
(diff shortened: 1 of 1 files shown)


troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v5.1.10

PR #368, opened 2021-06-01
fix: blinking on android when initialPage is used

ku8ar:
# Summary

This is more of showing how to fix the problem than the actual fix (unfortunately I couldn't build the code because I don't use typescript)

Currently, when we use initialPage on android, the change of the initial page is done with JS and requestAnimationFrame, which delays the call. When the pager-view is embedded in the navigation, this causes an initial blinking (or showing the first page briefly).
Additionally, initialPage causes a double call to onPageSelected (first with page = 0, then with page = initalPage).

This solution slightly changes the behavior of events. I'm delaying adding callbacks with evants here so that onPageSelected is not called initially.
I also puted setting initalPage straight to the native module so that delay is insignificant (and not affected by animations).

## What are the steps to reproduce (after prerequisites)?

1. create project with pager-view and react-navigation, add some stack with animations between screens
2. add pager-view on second screen and add any initalPage
3. check what happen when screen with pager-view is opened (doubled onPageSelected event and blinking)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
+++ b/android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
@@ -9,6 +9,8 @@ import java.util.*
 
 class FragmentAdapter(fragmentActivity: FragmentActivity) : FragmentStateAdapter(fragmentActivity) {
   private val childrenViews: MutableList<View> = ArrayList()
+  var initialPageinited = false
+
   override fun createFragment(position: Int): Fragment {
     return ViewPagerFragment(childrenViews[position])
   }
--- a/android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
+++ b/android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
@@ -32,31 +32,34 @@ class PagerViewViewManager : ViewGroupManager<ViewPager2>() {
     //https://github.com/callstack/react-native-viewpager/issues/183
     vp.isSaveEnabled = false
     eventDispatcher = reactContext.getNativeModule(UIManagerModule::class.java)!!.eventDispatcher
-    vp.registerOnPageChangeCallback(object : OnPageChangeCallback() {
-      override fun onPageScrolled(position: Int, positionOffset: Float, positionOffsetPixels: Int) {
-        super.onPageScrolled(position, positionOffset, positionOffsetPixels)
-        eventDispatcher.dispatchEvent(
-          PageScrollEvent(vp.id, position, positionOffset))
-      }
 
-      override fun onPageSelected(position: Int) {
-        super.onPageSelected(position)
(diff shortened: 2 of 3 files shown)


android/src/main/java/com/reactnativepagerview/FragmentAdapter.kt
```diff
@@ -9,6 +9,8 @@ import java.util.*
 
 class FragmentAdapter(fragmentActivity: FragmentActivity) : FragmentStateAdapter(fragmentActivity) {
   private val childrenViews: MutableList<View> = ArrayList()
+  var initialPageinited = false
```
ku8ar, line 12: I needed a local state to avoid calling setInitialPage every time a component render.
Can it be put in a better place?

android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
```diff
@@ -32,31 +32,34 @@ class PagerViewViewManager : ViewGroupManager<ViewPager2>() {
...
-        eventDispatcher.dispatchEvent(
-          PageScrollEvent(vp.id, position, positionOffset))
-      }
 
-      override fun onPageSelected(position: Int) {
-        super.onPageSelected(position)
-        eventDispatcher.dispatchEvent(
-          PageSelectedEvent(vp.id, position))
-      }
+    vp.post {
```
ku8ar, line 36: Here we delay adding events so as not to call onPageSelected, which always has position = 0 when creating a new viewpager

android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
```diff
@@ -200,6 +203,17 @@ class PagerViewViewManager : ViewGroupManager<ViewPager2>() {
     }
   }
 
+
+  @ReactProp(name = "initialPage", defaultInt = 0)
+  fun setInitialPage(viewPager: ViewPager2, value: Int) {
+    viewPager.post {
```
ku8ar, line 209: https://stackoverflow.com/a/60933258

android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
troZee: Could you also fix it on iOS ? 
Simulator Screen Shot - iPhone 11 - 2021-06-07 at 16 56 55

troZee (changes requested):
Thank you @ku8ar for your 1st contribution 🎉  I left some comments.

rashidul0405:
@troZee any chance of getting this merge soon?

ku8ar:
cls

Issue #323, opened 2021-04-08
KotlinNullPointerException


venkatpantham-mi:
## Environment
"react-native-pager-view": "^5.1.2"
OnePlus 8T
Android 11 (SDK 30)

## Issue
`com.reactnativepagerview.PagerViewViewManager.createViewInstance`

```
kotlin.KotlinNullPointerException: 
  at com.reactnativepagerview.PagerViewViewManager.createViewInstance (PagerViewViewManager.kt:30)
  at com.reactnativepagerview.PagerViewViewManager.createViewInstance (PagerViewViewManager.kt:21)
  at com.facebook.react.uimanager.ViewManager.createViewInstance (ViewManager.java:139)
  at com.facebook.react.uimanager.ViewManager.createView (ViewManager.java:82)
  at com.facebook.react.uimanager.NativeViewHierarchyManager.createView (NativeViewHierarchyManager.java:269)
  at com.facebook.react.uimanager.UIViewOperationQueue$CreateViewOperation.execute (UIViewOperationQueue.java:179)
  at com.facebook.react.uimanager.UIViewOperationQueue$1.run (UIViewOperationQueue.java:773)
  at com.facebook.react.uimanager.UIViewOperationQueue.flushPendingBatches (UIViewOperationQueue.java:888)
  at com.facebook.react.uimanager.UIViewOperationQueue.access$2200 (UIViewOperationQueue.java:42)
  at com.facebook.react.uimanager.UIViewOperationQueue$2.runGuarded (UIViewOperationQueue.java:846)
  at com.facebook.react.bridge.GuardedRunnable.run (GuardedRunnable.java:29)
  at android.os.Handler.handleCallback (Handler.java:883)
  at android.os.Handler.dispatchMessage (Handler.java:100)
  at android.os.Looper.loop (Looper.java:214)
  at android.app.ActivityThread.main (ActivityThread.java:7710)
  at java.lang.reflect.Method.invoke (Native Method)
  at com.android.internal.os.RuntimeInit$MethodAndArgsCaller.run (RuntimeInit.java:516)
  at com.android.internal.os.ZygoteInit.main (ZygoteInit.java:950)

```  
  This is the error I'm getting from the google play console

efstathiosntonas:
Same here, it occured in several SDKs from 7 up to 11 and in several devices

troZee:
Hello 👋 
[...]

PR #369, opened 2021-06-07
[issue-323] Fix KotlinNullPointerException 

zabutok:
# Summary
Fragment Adapter use currentActivity before it was initiated so we add LifecycleEventListener to check when app resume work fix#[323](https://github.com/callstack/react-native-pager-view/issues/323)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌    |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
+++ b/android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
@@ -7,6 +7,7 @@ import androidx.viewpager2.widget.ViewPager2.OnPageChangeCallback
 import com.facebook.infer.annotation.Assertions
 import com.facebook.react.bridge.ReadableArray
 import com.facebook.react.common.MapBuilder
+import com.facebook.react.bridge.LifecycleEventListener
 import com.facebook.react.uimanager.PixelUtil
 import com.facebook.react.uimanager.ThemedReactContext
 import com.facebook.react.uimanager.UIManagerModule
@@ -25,10 +26,24 @@ class PagerViewViewManager : ViewGroupManager<ViewPager2>() {
     return REACT_CLASS
   }
 
+  fun createFragmentAdapter(reactContext: ThemedReactContext): FragmentAdapter {
+    return FragmentAdapter((reactContext.currentActivity as FragmentActivity?)!!)
+  }
+
   override fun createViewInstance(reactContext: ThemedReactContext): ViewPager2 {
     val vp = ViewPager2(reactContext)
-    val adapter = FragmentAdapter((reactContext.currentActivity as FragmentActivity?)!!)
-    vp.adapter = adapter
+    if(reactContext.hasCurrentActivity()) {
+      vp.adapter = createFragmentAdapter(reactContext);
+    }
+    reactContext.addLifecycleEventListener(object : LifecycleEventListener {
+      override fun onHostResume() {
+        if(vp.adapter == null && reactContext.hasCurrentActivity()) {
+          vp.adapter = createFragmentAdapter(reactContext)
+        }
+      }
+      override fun onHostPause() {}
+      override fun onHostDestroy() {}
+    });
     //https://github.com/callstack/react-native-viewpager/issues/183
     vp.isSaveEnabled = false
     eventDispatcher = reactContext.getNativeModule(UIManagerModule::class.java)!!.eventDispatcher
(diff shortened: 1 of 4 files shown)


troZee (changes requested):
Hey @zabutok 👋 
Thank you for your 1st contribution 🎉  Could you convert your solution a little bit using this approach ? 

https://reactnative.dev/docs/native-modules-android#listening-to-lifecycle-events

So it should like like below: 

https://stackoverflow.com/a/52516893/4247480

android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
```diff
@@ -25,10 +26,24 @@ class PagerViewViewManager : ViewGroupManager<ViewPager2>() {
...
+  }
+
   override fun createViewInstance(reactContext: ThemedReactContext): ViewPager2 {
     val vp = ViewPager2(reactContext)
-    val adapter = FragmentAdapter((reactContext.currentActivity as FragmentActivity?)!!)
-    vp.adapter = adapter
+    if(reactContext.hasCurrentActivity()) {
+      vp.adapter = createFragmentAdapter(reactContext);
+    }
+    reactContext.addLifecycleEventListener(object : LifecycleEventListener {
```
troZee, line 38: Please also remove the listener from reactContext instance to avoid memory leaks:
```
  override fun onDropViewInstance(view: ViewPager2) {
    super.onDropViewInstance(view)
    UIManagerHelper.getReactContext(view).removeLifecycleEventListener(this);
  }
 ```

**Review by troZee — CHANGES_REQUESTED** [review]

**Comment by troZee** [comment]
@zabutok Could you also provide a testing plan before I merge it ? 
I am wondering about the below case: 
- Activity is not created yet 
- View has been created 
- View Manager calls `addViews` method 
- Child is not added, bc adapter is still null 
- Adapter is initialized in  `onHostResume` method 

and the question is, what about in above case ? is this case would occur on the native side ?

**Comment by zabutok** [comment]
> @zabutok Could you also provide a testing plan before I merge it ?
> I am wondering about the below case:
> 
> * Activity is not created yet
> * View has been created
> * View Manager calls `addViews` method
> * Child is not added, bc adapter is still null
> * Adapter is initialized in  `onHostResume` method
> 
> and the question is, what about in above case ? is this case would occur on the native side ?

@troZee it is good question how i can understood addViews to detached ViewGroupManager it is not realistic steps(cos we don't have connected views currentActivity empty cos we fetching js bundle) if i am wrong we will have empty tab view i will try reproduce this behavior

**Comment by troZee** [comment]
> > @zabutok Could you also provide a testing plan before I merge it ?
> > I am wondering about the below case:
> > 
> > * Activity is not created yet
> > * View has been created
> > * View Manager calls `addViews` method
> > * Child is not added, bc adapter is still null
> > * Adapter is initialized in  `onHostResume` method
> > 
> > and the question is, what about in above case ? is this case would occur on the native side ?
> 
> @troZee it is good question how i can understood addViews to detached ViewGroupManager it is not realistic steps(cos we don't have connected views currentActivity empty cos we fetching js bundle) if i am wrong we will have empty tab view i will try reproduce this behavior

I am not sure, if above case is reproducible and it was just only my concerns. We can try for now your solution, but in the nearest future, we need to find a way, to create an Adapter without Activity to avoid above NPE.

File: android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt [hunk 2]
```diff
@@ -25,10 +27,25 @@ class PagerViewViewManager : ViewGroupManager<ViewPager2>() {
...
+  }
+
   override fun createViewInstance(reactContext: ThemedReactContext): ViewPager2 {
     val vp = ViewPager2(reactContext)
-    val adapter = FragmentAdapter((reactContext.currentActivity as FragmentActivity?)!!)
-    vp.adapter = adapter
+    if(reactContext.hasCurrentActivity()) {
+      vp.adapter = createFragmentAdapter(reactContext);
+    }
+    lifecycleEventListener = object : LifecycleEventListener {
```
zabutok, line 41: I wanted to do as you extended, but then I made that without a closure, we don`t have access to reactContext and ViewPager, what do you think about this solution?

troZee (approved):

troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v5.2.1

PR #376, opened 2021-06-14
Render blocks UI thread

alpha0010:
# Summary

The LazyPager render trigger can block UI https://github.com/callstack/react-native-pager-view/issues/104#issuecomment-860518078.  This PR aims to resolve such by changing event scheduling.

### What's required for testing (prerequisites)?

"Lazy Render Example" screen from the example app.

### What are the steps to reproduce (after prerequisites)?

Navigating between pages should work the same as current `next` branch (firing events, mount/unmount pages).  I have not observed the issue locally; maybe @DanijelBojcic can comment?

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [x] I have tested this on a device and a simulator
- [ ] I added the documentation in `README.md`
- [ ] I updated the typed files (TS and Flow)

PR #380, opened 2021-06-18
fix: Render when scroll is idle

DanijelBojcic:
# Summary

This PR reschedules the new page render so it doesn't block the UI thread and the animation events. 

The bug is mentioned in this [comment](https://github.com/callstack/react-native-pager-view/issues/104#issuecomment-860518078), and the fix is based on #376  PR. 
Thanks to @alpha0010 !

Instead of rendering on the `onPageSelected` event, now the render starts if the `scrollState` is `idle`.

Run the example project, it should work the same.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist


src/LazyPagerView.tsx
```diff
@@ -221,14 +232,38 @@ class LazyPagerViewImpl<ItemT> extends React.Component<
...
+    }
   };
 
   private onPageSelected = (event: PagerViewOnPageSelectedEvent) => {
     const currentPage = event.nativeEvent.position;
 
     // Ignore spurious events that can occur on mount with `initialPage`.
     // TODO: Is there a way to avoid triggering the events at all?
-    if (this.isNavigatingToPage !== null) {
+    if (this.isNavigatingToPage != null) {
```
troZee, line 266: Could you explain, why it has been changed ?

troZee:
@alpha0010 could you help me to check this PR ?

lib/commonjs/LazyPagerView.js
```diff
@@ -0,0 +1,314 @@
+"use strict";
+
+Object.defineProperty(exports, "__esModule", {
```
troZee, line 3: please revert this file

alpha0010:
I have not yet had a chance to test.  One aspect to check: is the `idle` event guaranteed to fire?  Specifically, if programmatically setting pages without animation, or using curl transition?

DanijelBojcic:
> I have not yet had a chance to test. One aspect to check: is the `idle` event guaranteed to fire? Specifically, if programmatically setting pages without animation, or using curl transition?

I think you are right. Also I just realized that state change doesn't always means that the selected page was changed, so its doing unneeded calls.

DanijelBojcic:
Maybe we should implement reanimated. What are your thoughts?

alpha0010:
I do not think it would be good to add a dependency on reanimated, especially since it is a more complicated library to add to an app.

An alternative solution could be implementing on your app side.
- Make pages lightweight shells (placeholder graphics?)
- On page mount, listen to events from pagerview
- Swap the content of a placeholder page with the heavy real content when either condition becomes true:
  - The page is the currently selected page
  - It has been *X* milliseconds (randomized so not all pages trigger at once) since the last pagerview event

(If you have not already, you could play around with `buffer` and `maxRenderWindow` to see if you can find a compromise that reduces blocks, but still has enough preloaded for swiping.)

troZee:
What should we do with it ?

DanijelBojcic:
I'll close this PR as it does not work as expected.

PR #397, opened 2021-07-21
fix(android): implement getContext for the older versions of react native

### LINKED ISSUE #394: [Android] getReactContext does not work in < 63
## Environment

"react": "16.9.0",
"react-native": "0.61.5",
"react-native-pager-view": "^5.2.1",

## Description

The error was the following one:

Captura de Tela 2021-07-20 às 18 34 11

I'm afraid of this recent version of react-native-pager-view only works with react-native 0.63+ because of the getReactContext() function.

## Reproducible Demo

To reproduce it is just run an android build, even in debug it shouldn't works. 
OBS: you need to use the environment above.

troZee:
# Summary
Closes #394 

`getReactContext` method is missing in the older versions of react native. Moved this [method](https://github.com/facebook/react-native/blob/v0.64.2/ReactAndroid/src/main/java/com/facebook/react/uimanager/UIManagerHelper.java#L138) to the codebase. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

### DIFF
--- a/android/src/main/java/com/reactnativepagerview/Helper.kt
+++ b/android/src/main/java/com/reactnativepagerview/Helper.kt
@@ -1,0 +1,21 @@
+package com.reactnativepagerview
+
+import android.content.Context
+import android.content.ContextWrapper
+import android.view.View
+import com.facebook.react.bridge.ReactContext
+
+
+class Helper {
+    companion object {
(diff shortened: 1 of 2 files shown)


troZee:
cc @mikeqcp

mikeqcp (approved):

PR #403, opened 2021-07-22
chore: remove lib folder

troZee:
lib folder should not be kept inside the repository, bc it is generated each time by bob. 
Additionally I updated the gitignore file


troZee:
cc @mikeqcp

mikeqcp (approved):

PR #404, opened 2021-07-23
refactor: use RecyclerView.Adapter instead of FragmentStateAdapter

### LINKED ISSUE #389: [Android] Sometimes the component render nothing (related to #323)
## Environment

platform: Android only
react-native: 0.62.3
react-native-pager-view: 5.2.1

## Description

After I upgrade to 5.2.1 (#323), I didn't see the NPE crash anymore. However, there is a side effect so that sometimes we see that the component just renders nothing.

In short, the crash is replaced with null rendering.

## Reproducible Demo

Good news is that I can now consistently reproduce this one. The step is actually simple:

1. run `npx react-native run-android` when your connected Android device's screen is **off**
2. After the command finishes, you may need to wait a few seconds, making sure that the the bundle is loaded properly.
3. Unlock your phone, the app should load with the PagerView component showing nothing. When this happens, the swipe effect is not working also.

mikeqcp:
# Summary

Closes #389

It rewrites ViewPager2 adapter class.
Currently it uses implementation of `FragmentStateAdapter` - I've changed it to use `RecyclerView.Adapter` implementation.
The main profit of this change is the fact that new adapter doesn't need `FragmentActivity` in it's constructor - it simplifies process of creating the viewpager instance.

It should #389 - this issue was happening because React activity was not initialised at the time of creating viewPager. New implementation doesn't' require activity to be initialised, so the issue doesn't happen.

Run any example on Android.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist


android/src/main/java/com/reactnativepagerview/ViewPagerAdapter.kt
```diff
@@ -0,0 +1,58 @@
...
+    notifyItemInserted(index)
+  }
+
+  fun getChildAt(index: Int): View {
+    return childrenViews[index]
+  }
+
+  fun removeChild(child: View) {
+    val index = childrenViews.indexOf(child)
+    removeChildAt(index)
```
troZee, line 47:
```suggestion
    removeChildAt(index)
    notifyItemRemoved(index)
```

android/src/main/java/com/reactnativepagerview/ViewPagerAdapter.kt
```diff
@@ -0,0 +1,58 @@
...
+  override fun onBindViewHolder(holder: ViewPagerViewHolder, index: Int) {
+    val container: FrameLayout = holder.container
+    val child = getChildAt(index)
+
+    if (container.childCount > 0) {
+      container.removeAllViews()
+    }
+
+    if (child.parent != null) {
+      (child.parent as ViewGroup).removeView(child)
```
troZee, line 26:
```suggestion
      (child.parent as FrameLayout).removeView(child)
```

android/src/main/java/com/reactnativepagerview/ViewPagerAdapter.kt
```diff
@@ -0,0 +1,58 @@
...
+    return childrenViews[index]
+  }
+
+  fun removeChild(child: View) {
+    val index = childrenViews.indexOf(child)
+    removeChildAt(index)
+  }
+
+  fun removeAll() {
+    childrenViews.clear()
```
troZee, line 51:
```suggestion
    childrenViews.clear()
    notifyXXX()
```

android/src/main/java/com/reactnativepagerview/ViewPagerViewHolder.kt
```diff
@@ -0,0 +1,20 @@
...
+
+
+class ViewPagerViewHolder private constructor(container: FrameLayout) : ViewHolder(container) {
+    val container: FrameLayout
+        get() = itemView as FrameLayout
+
+    companion object {
+        fun create(parent: ViewGroup): ViewPagerViewHolder {
+            val container = FrameLayout(parent.context)
+            container.layoutParams = ViewGroup.LayoutParams(-1, -1)
```
troZee, line 15: why -1 ? should not be `MATCH_PARENT` ?

troZee (changes requested):

android/src/main/java/com/reactnativepagerview/ViewPagerAdapter.kt
mikeqcp: don't need `notifyItemRemoved` here because it's already invoked inside `removeChildAt`

PR #407, opened 2021-07-26
fix(android): set initial page natively

### LINKED ISSUE #400: initial page render question
## Ask your Question

when PagerView rendered, 
even though the initialPage={1} is selected, Page 0 is displayed first then Page 1 is displayed.
i use Android simulator. help!

mikeqcp:
# Summary

Closes #400

On Android, setting initial page was implemented asynchronously on React side and native `onPageSelected` was fired twice (with default value and then with proper one).
This PR moves the implementation directly to `ViewPager2` API.

Setup `Basic Example` with `initialPage={1}` and with `onPageSelected={(e: any) => console.log(e.nativeEvent.position)}`

Open the Example and see what pages are reported in `onPageSelected` callback

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
+++ b/android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
@@ -52,31 +52,37 @@ class PagerViewViewManager : ViewGroupManager<ViewPager2>() {
     //https://github.com/callstack/react-native-viewpager/issues/183
     vp.isSaveEnabled = false
     eventDispatcher = reactContext.getNativeModule(UIManagerModule::class.java)!!.eventDispatcher
-    vp.registerOnPageChangeCallback(object : OnPageChangeCallback() {
-      override fun onPageScrolled(position: Int, positionOffset: Float, positionOffsetPixels: Int) {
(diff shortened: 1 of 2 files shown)


android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
```diff
@@ -139,6 +145,13 @@ class PagerViewViewManager : ViewGroupManager<ViewPager2>() {
     viewPager.isUserInputEnabled = value
   }
 
+  @ReactProp(name = "initialPage", defaultInt = 0)
+  fun setInitialPage(viewPager: ViewPager2, value: Int) {
+    viewPager.post {
```
mikeqcp, line 150: It needs to be set within `post` because `ViewPager2` has an issue and initial page doesnt change unless this call is defered to the next cycle
more details here: https://stackoverflow.com/questions/56311862/viewpager2-default-position

troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v5.3.3

PR #410, opened 2021-07-28
fix #409

### LINKED ISSUE #409: Blink initial page was solved by 5.4.0, but the initial page is not rendered on Android 6.0
## Environment

react native 0.63.4
Android 6.0.0 API 23

## Description

I upgrade the package from 5.2.4 to 5.4.0, because of the blink initial page. It works well on my phone - Android 10.
But when I use my APP on the Android 6.0 device, the initial page is not rendered. After I scroll the page, the component works well.

JackClown:
# Summary

Fix issue #409, although I think it's not a good solution. I just need help to solve the problem

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅    |

## Checklist


troZee:
Hey @JackClown 
I have converted it into a draft, bc it is a temporary workaround for setting initial index.  Probably pager tries to set page, once it has not been fully initialized. I need to take a look on it.

troZee:
@JackClown is it still an issue ?

PR #412, opened 2021-07-29
fix setPage call to long

NiuGuohui:
### *It's too long when call setPage method if is running other animation.*

# Summary
It's too long when call `setPage` method if is running other animation. I thought its waiting for other animation frame. So, I tried change`requestAnimationFrame` to `setTimeout`.
It looks smoother.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅   |

## Checklist

### DIFF
--- a/src/LazyPagerView.tsx
+++ b/src/LazyPagerView.tsx
@@ -154,7 +154,7 @@ class LazyPagerViewImpl<ItemT> extends React.Component<
       })
     );
     // Send paging command.
-    setTimeout(() => {
+    setImmediate(() => {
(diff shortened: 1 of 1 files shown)


troZee:
@alpha0010 WDYT ?

alpha0010:
`setImmediate()` might be more appropriate, but yes, seems fine.

NiuGuohui:
> `setImmediate()` might be more appropriate, but yes, seems fine.

Yes,I tried it.`setImmediate()` is more appropriate.

troZee (approved):

PR #474, opened 2021-11-04
Add some extreme cases to test

yinjun322:
# Add some extreme cases to test

1. Removing many pages at once can cause  crash
2. Quick page jumping, a blank page will appear.(eg: 1 to 20)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ❌     |

## Checklist


yinjun322:
I found that the completion callback is not executed when setViewControllers:action:animated:completion is called multiple times. And UIPageViewController navigates to blank page.

PR #476, opened 2021-11-05
Create README.md

dbi1:
# Summary

Simple instructions on how to run the example app.

Tested on both simulators (iOS and Android), the app launches fine in both cases when following these instructions.

Thank you for considering this PR.

### DIFF
--- a/example/README.md
+++ b/example/README.md
@@ -1,0 +1,9 @@
+# Run the Example App
+
+Clone the repo, then (inside this directory):
+
+    yarn install
+    cd ios && pod install && cd ..
+    react-native start
+    react-native run-android
+    react-native run-ios


example/README.md
```diff
@@ -0,0 +1,9 @@
+# Run the Example App
```
troZee, line 1: All info is included here: https://github.com/callstack/react-native-pager-view/blob/master/CONTRIBUTING.md

dbi1: Thank you @troZee.

Maybe a link to this file from the above README.md might be a good idea then, instead? It would not have occurred to me to look into CONTRIBUTING.md, as I was just trying to see the PagerView in action...

Issue #315, opened 2021-03-31
[iOS] go back gesture does not work


KingAmo:
## Description

Go back gesture is blocked, bc all events are captured by viewpager, hence react navigation is not aware of any pan event. 

![image](https://user-images.githubusercontent.com/12766071/126637912-64b4df4c-553d-4a1a-9d28-236fdf9fa938.png)

## Reproducible Demo

Go to basic example and try to use go back gesture

troZee:
Together with @mikeqcp were trying to find any viable solution. Below I am providing a result of our investigation.

- When we set `self.scrollView.scrollEnabled = NO` to `UIScrollView`, it works fine, bc events are captured by react navigation. But then we cannot enable the scrolling inside the pagerview, bc pagerview would not capture any events. 

- We were trying to implement `touchesBegan/touchesEnded/touchesMoved` but it did not changed anything

svenlombaert:
Same issue when using `react-native-navigation`. I'm using https://github.com/satya164/react-native-tab-view and it indeed seems to capture all gestures, can't swipe to open side menu as well.

happyfloat:
In my old version I did use [react-native-best-viewpager](https://github.com/abbasmoosavi/react-native-best-viewpager) and both worked well, not sure if the foundation is comparable, but maybe you can check it out.

MonsterAnan:
+1 
Same issue when using react-native-navigation.

JoniVR:
FYI, for `react-navigation` (and perhaps also `react-native-navigation`):
downgrading `react-native-tab-view` to `v2.16.0` seems to resolve the issue for now.

1061750360:
I have a temporary solution. When you want to go back, you should forbid ` react-native-pager-view`,let it can not scroll.
(My English is not good, so Please understand)

This is the code:

step 1:
pagerView has three child, curTabIndex is current child index. So **curTabIndex === 0** is the first child,And now you wan t to go back. And the **evt.nativeEvent.pageX <= 50** is true(because IOS back gesture distance Approximately equal to
50).you forbid ` react-native-pager-view`,let it can not scroll.
**Now you can use gesture to go back!**
```
  const panResponder = useRef(
      PanResponder.create({
        onMoveShouldSetPanResponder: () => true,
        onPanResponderGrant: (evt, gestureState) => {
          if (evt.nativeEvent.pageX <= 50 && curTabIndex === 0) {
            pagerViewRef.current.setScrollEnabled(false)
          }
[...]

PR #500, opened 2022-01-16
fix: disable pager scroll gesture on swipe to go back

intergalacticspacehighway:
# Summary

Fixes  https://github.com/callstack/react-native-pager-view/issues/315

- This PR adds a gesture failure logic on scroll view used by pager if pop gesture is triggered on iOS.
- It checks if a navigation controller is present and adds [requireGestureRecognizorToFail](https://stackoverflow.com/questions/21928023/may-i-know-what-exactly-requiregesturerecognizertofail-will-do) on ScrollView used by Pager view.

## Test Plan

- Created a [repo](https://github.com/intergalacticspacehighway/pager-view-ios-swipe-back-fix) with the fix. App.js includes react navigation code with pager view in `ProfileScreen`
- Test the app with and without react navigation. The swipe to go back gesture should work as expected
https://user-images.githubusercontent.com/23293248/149651316-e1e7eafe-7d03-4204-b9ef-eddb6e1bb873.mov

- Apply the below patch in react-native-pager-view
- Test in a RN app with react navigation and try swipe to go back gesture on pager view
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅  |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -89,6 +89,10 @@
         [self embed];
         [self setupInitialController];
     }
+
+    if (self.reactViewController.navigationController != nil && self.reactViewController.navigationController.interactivePopGestureRecognizer != nil) {
+        [self.scrollView.panGestureRecognizer requireGestureRecognizerToFail:self.reactViewController.navigationController.interactivePopGestureRecognizer];
+    }
 }
 
 - (void)embed {
@@ -167,11 +171,6 @@
                             direction:UIPageViewControllerNavigationDirectionForward
                              animated:YES
              shouldCallOnPageSelected:YES];
-
-              
-        if (self.reactViewController.navigationController != nil && self.reactViewController.navigationController.interactivePopGestureRecognizer != nil) {
-            [self.scrollView.panGestureRecognizer requireGestureRecognizerToFail:self.reactViewController.navigationController.interactivePopGestureRecognizer];
-        }
     }
 }
 
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -167,6 +167,11 @@
                             direction:UIPageViewControllerNavigationDirectionForward
                              animated:YES
              shouldCallOnPageSelected:YES];
+
+              
+        if (self.reactViewController.navigationController != nil && self.reactViewController.navigationController.interactivePopGestureRecognizer != nil) {
+            [self.scrollView.panGestureRecognizer requireGestureRecognizerToFail:self.reactViewController.navigationController.interactivePopGestureRecognizer];
+        }
     }
 }
 


intergalacticspacehighway:
While doing more testing I found that `self.reactViewController.navigationController` can be nil sometimes. Not sure why. When it's not nil, we get the desired result. I guess if we can get the `navigationController` each time, we can solve this issue. Let me know if someone finds it. I'll keep on looking too!

arunim2405:
@intergalacticspacehighway I think it takes some time for self.reactViewController to be initialized and setupInitalController can finish before that. Is it possible to introduce a small timeout in this or call a function when self.reactViewController is changed.
I added this code to scrollViewWillBeginDragging and it fails the first drag gesture but works after.

rkritskiy:
@intergalacticspacehighway @arunim2405 hi, did you manage to solve this problem in any way?

intergalacticspacehighway:
@rkritskiy @arunim2405 moving the block in `didMoveToWindow` works for me. Just pushed the [changes](https://github.com/callstack/react-native-pager-view/pull/500/commits/3ad752ae302c407b3a3a2b3f33a74166630f7e28). Can you verify?

rkritskiy:
@intergalacticspacehighway I have the fullScreenGestureEnabled setting for the navigator and it looks like swipe back doesn't work with this change

hirbod:
I can confirm this works. Will anybody merge this? :)

raymondjacobson:
I pulled in this patch, but still some amount of the time, the pager view is stealing swipes backwards from the stack navigator. Not sure if I'm doing anything wrong. @hirbod do you have a link to the repo where this worked for you?

raymondjacobson:
> @intergalacticspacehighway I have the fullScreenGestureEnabled setting for the navigator and it looks like swipe back doesn't work with this change

Hey @rkritskiy, I managed to get this working. Will try to put up some code, but here's a vid, and generally what I did:

In RNSStackScreen, `setupGestureHandlers`

Instead of adding the panRecognizer to `self`, I did this instead
```
[_controller.view addGestureRecognizer:_panRecognizer];
```

This let's me get it in the PageView like this:
```
self.reactViewController.navigationController.view.gestureRecognizers
```
(I'm no iOS dev, so there may be a way better way to do this).

In RNSStackScreen, `gestureRecognizerShouldBegin`

Disable the pan gesture if we are swiping right (I only care about swiping backwards).
```
if (topScreen.fullScreenSwipeEnabled) {
    // we want only `RNSPanGestureRecognizer` to be able to recognize when
    // `fullScreenSwipeEnabled` is set
    if ([gestureRecognizer isKindOfClass:[RNSPanGestureRecognizer class]]) {
        RNSPanGestureRecognizer *pan = gestureRecognizer;
      _isFullWidthSwiping = YES;
      [self cancelTouchesInParent];
        CGPoint translation = [pan translationInView:_controller.view];
        if (translation.x < 0) {
            return NO;
        }
      return YES;
    }
    return NO;
  }
```

In ReactNativePageView, `didMoveToWindow`, change to

```
if (self.reactViewController.navigationController != nil) {
        for (UIGestureRecognizer *recognizer in self.reactViewController.navigationController.view.gestureRecognizers) {
             [self.scrollView.panGestureRecognizer requireGestureRecognizerToFail:recognizer];
        }
    }
```

(this is blunt, maybe a better way?)

And finally in ReactNativePageView, `didFinishAnimating`

```
if (self.reactViewController.navigationController != nil) {
            if (currentIndex == 0) {
                for (UIGestureRecognizer *recognizer in self.reactViewController.navigationController.view.gestureRecognizers) {
                    recognizer.enabled = YES;
                     [self.scrollView.panGestureRecognizer requireGestureRecognizerToFail:recognizer];
                }
            } else {
                for (UIGestureRecognizer *recognizer in self.reactViewController.navigationController.view.gestureRecognizers) {
                    recognizer.enabled = NO;
                     [self.scrollView.panGestureRecognizer requireGestureRecognizerToFail:NULL];
                }
            }
        }
```


https://user-images.githubusercontent.com/2731362/162644941-0a936e9b-7a15-4a79-bf4f-d2bbc777d1d5.mov

intergalacticspacehighway:
@rkritskiy yeah noticed `fullScreenGestureEnabled` prevents the default `interactivePopGestureRecognizer` [here by returning NO](https://github.com/software-mansion/react-native-screens/blob/main/ios/RNSScreenStack.m#L627). This is a bit tricky. 

@raymondjacobson this is cool! Feels weird to mutate the navigation controller's gesture recognizer though. `recognizer.enabled = NO;`. This might require adding some cleanup logic to reset them (in case the component unmounts)

The solution in PR works without `fullScreenGestureEnabled` [demo](https://github.com/intergalacticspacehighway/pager-view-ios-swipe-back-fix). Maybe we can do a cleaner solution to support `fullScreenGestureEnabled` using gesture recognizer delegate 🤔. Will post it here once I have another solution!

raymondjacobson:
> @rkritskiy yeah noticed `fullScreenGestureEnabled` prevents the default `interactivePopGestureRecognizer` [here by returning NO](https://github.com/software-mansion/react-native-screens/blob/main/ios/RNSScreenStack.m#L627). This is a bit tricky.
> 
> @raymondjacobson this is cool! Feels weird to mutate the navigation controller's gesture recognizer though. `recognizer.enabled = NO;`. This might require adding some cleanup logic to reset them (in case the component unmounts)
> 
> The solution in PR works without `fullScreenGestureEnabled` [demo](https://github.com/intergalacticspacehighway/pager-view-ios-swipe-back-fix). Maybe we can do a cleaner solution to support `fullScreenGestureEnabled` using gesture recognizer delegate 🤔. Will post it here once I have another solution!

1000000%. I'm no iOS dev! "It works" was all I was going for in that comment :)

raymondjacobson:
Modified the "demo" in case that's at all useful. https://github.com/raymondjacobson/pager-view-ios-swipe-back-fix/commit/1319d91f74da94532ba327fb92a69dc83090f5c9

Aryk:
I've historically had problems applying changes to native code with Expo. For any expo people out there, would the process, simply be:

1. Add [this patch](https://github.com/callstack/react-native-pager-view/pull/500/files) to patches folder 
2. Use patch-package to apply patch
3. Recompile the app using EAS

@hirbod do you think this will work?

intergalacticspacehighway:
The PR works pretty well without `fullScreenSwipeEnabled` and the solution looks clean to me. 

But, to support `fullScreenSwipeEnabled` I've found another approach (it requires a patch in RN screens 😬 )

This solution is derived by referring apps that have a full-screen swipe to go back gesture. (Twitter and Instagram).

### Twitter and Instagram do below when they have tabs. 
- Swipe to go back with `screen edge gesture` works on each tab. Full Screen swipe to go back works only on first tab.

To achieve the above we'll need [this patch in RNSScreenStack](https://github.com/intergalacticspacehighway/pager-view-ios-swipe-back-fix/commit/50d5bb979cba630ff00f3b6fac242ce0e4bb002b) along with the change in this PR. 


https://user-images.githubusercontent.com/23293248/166097551-88c3bca7-0c11-400b-b797-e47a9a305466.mov

intergalacticspacehighway:
One more way to achieve this would be if we can attach a delegate to the gesture recognizer of the scroll view used by the pager view and accept simultaneous gesture with the RNScreens gesture, I tried a few approaches but was mostly unsuccessful as I couldn't find a way to override the pager-view's ScrollView or its gesture delegate 😐

hirbod:
There is an ongoing PR at react native screens, let me grab that for you

giangpro93:
I tested it with @react-navigation/material-top-tabs and it doesn't seem to work. Not sure if is because of their wrapper layer or if this still doesn't work.

intergalacticspacehighway:
@giangpro93 can you share a repo with a minimal example? I can take a look

giangpro93:
That is awesome. I will try to do that tomorrow and let you know.

troZee:
Hey @intergalacticspacehighway 
TBH I don't understand, where exactly the issue is. I have checked the example app from the master branch and the swipe to  go back gesture works fine.  

I think, we should distinguish two things: 

1. go back swipe behaviour from the screen edge (Supported by native apps (App Store))
2. go back swipe behaviour from the any point in the screen (Not supported by native apps) 

Ad 1. 
Currently pager example supports a such behaviour, so no fixes are needed 

Ad.2 
This is a custom solution and this behaviour is not a native solution. If we would like to implement a such feature, please create a separate issue, where all feature description will be provided. 

Could you clarify, why this fix is needed ? Maybe I don't understand the issue. 

https://user-images.githubusercontent.com/12766071/166911998-751356c6-4ca2-408d-a2fd-393e87787fab.mov
https://user-images.githubusercontent.com/12766071/166912040-70761183-384a-43de-87a7-787f1c21c15e.mov

hirbod:
I just added both patches by @intergalacticspacehighway and also updated them for more recent versions, working great for me for now.

```diff
diff --git a/node_modules/react-native-pager-view/ios/ReactNativePageView.m b/node_modules/react-native-pager-view/ios/ReactNativePageView.m
index 8f20bf8..9ec716f 100644
--- a/node_modules/react-native-pager-view/ios/ReactNativePageView.m
+++ b/node_modules/react-native-pager-view/ios/ReactNativePageView.m
@@ -89,6 +89,9 @@ - (void)didMoveToWindow {
         [self embed];
         [self setupInitialController];
     }
+    if (self.reactViewController.navigationController != nil && self.reactViewController.navigationController.interactivePopGestureRecognizer != nil) {
+        [self.scrollView.panGestureRecognizer requireGestureRecognizerToFail:self.reactViewController.navigationController.interactivePopGestureRecognizer];
+    } 
 }
 
 - (void)embed {
```

```diff
diff --git a/node_modules/react-native-screens/ios/RNSScreenStack.m b/node_modules/react-native-screens/ios/RNSScreenStack.m
index 47c8f8d..d261d0e 100644
--- a/node_modules/react-native-screens/ios/RNSScreenStack.m
+++ b/node_modules/react-native-screens/ios/RNSScreenStack.m
@@ -10,6 +10,7 @@
 #import <React/RCTTouchHandler.h>
 #import <React/RCTUIManager.h>
 #import <React/RCTUIManagerUtils.h>
+#import "ReactNativePageView.h"
 
 @interface RNSScreenStackView () <
     UINavigationControllerDelegate,
@@ -594,6 +595,10 @@ - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
     return NO;
   }
 
+  if ([gestureRecognizer isKindOfClass:[_controller.interactivePopGestureRecognizer class]]) {
+    return YES;
+  }
+
 #if TARGET_OS_TV
   [self cancelTouchesInParent];
   return YES;
@@ -637,6 +642,20 @@ - (BOOL)gestureRecognizerShouldBegin:(UIGestureRecognizer *)gestureRecognizer
 #endif
 }
 
+- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecognizeSimultaneouslyWithGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer {
+    if ([otherGestureRecognizer isKindOfClass: NSClassFromString(@"UIScrollViewPanGestureRecognizer")] && [otherGestureRecognizer.view.reactViewController isKindOfClass: [UIPageViewController class]]) {
+      UIPageViewController* pageController = otherGestureRecognizer.view.reactViewController;
+      if (pageController != nil && [pageController.delegate isKindOfClass:[ReactNativePageView class]]) {
+        ReactNativePageView* page = pageController.delegate;
+        if (page != nil && page.currentIndex == 0) {
+          return YES;
+        }
+      }
+    }
+    return NO;
+}
+
+
 #if !TARGET_OS_TV
 - (void)setupGestureHandlers
 {
```

andreialecu:
@troZee the swipe to go back gesture does indeed stop working *sometimes* in the pager view. Most times it works, but randomly it will stop working.

Previously with `react-native-collapsible-tab-view` we were using a horizontal scroll view, which coincidentally behaved exactly the same (hit and miss about when it would let the gesture thorough). See: https://github.com/PedroBern/react-native-collapsible-tab-view/issues/248#issuecomment-1081110155

We've been able to work around it by setting `hitSlop={left: -10}` on the scroll view. That way touches within 10 pixels from the edge would just be passed through and it was enough to fix this.

PagerView does not seem to expose a `hitSlop` property though.

andreialecu:
Oh, it appears like adding a wrapper view with the hitslop fixes this:

```tsx
    <View hitSlop={{left: -10}}>
		<PagerView...
```

hirbod:
I prefer the patch since this works just natively and especially with the fullscreen swipe gesture it does behave like Twitter

troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v5.4.19

ramiel1999:
updating react-native-page-view to latest version 5.14.24 and still can not swipe back

Issue #440, opened 2021-09-13
Crash when navigating back from PagerView


alexco2:
Hello, I found a crash in my app which I was able to reproduce and trace back to `PagerView`.

## Environment

    "@react-navigation/bottom-tabs": "^6.0.0",
    "@react-navigation/native": "^6.0.0",
    "@react-navigation/native-stack": "^6.0.0",
    "react-native-pager-view": "^5.4.1",
    "react-native-screens": "~3.4.0",

## Description

This bug crashed the app without an error message. Crashlytics gives me the following crashreport:

```
Fatal Exception: java.lang.IllegalArgumentException: Scrapped or attached views may not be recycled. isScrap:false isAttached:true b.v.b.f$m{8268e2 VFED..... ......ID 0,0-1080,2096 #3}, adapter:com.reactnativepagerview.b@447b473, layout:b.v.b.f$h@f750f30, context:com.facebook.react.uimanager.l0@5d374a0
       at androidx.recyclerview.widget.RecyclerView$v.C(:282)
       at androidx.recyclerview.widget.RecyclerView.removeAnimatingView(:22)
       at androidx.recyclerview.widget.RecyclerView$m.a(:27)
       at androidx.recyclerview.widget.RecyclerView$l.h(:7)
       at androidx.recyclerview.widget.n.G(:3)
       at androidx.recyclerview.widget.c$d.onAnimationEnd(:17)
       at android.view.ViewPropertyAnimator$AnimatorEventListener.onAnimationEnd(ViewPropertyAnimator.java:1111)
       at android.animation.Animator$AnimatorListener.onAnimationEnd(Animator.java:554)
       at android.animation.ValueAnimator.endAnimation(ValueAnimator.java:1250)
       at android.animation.ValueAnimator.doAnimationFrame(ValueAnimator.java:1492)
       at android.animation.AnimationHandler.doAnimationFrame(AnimationHandler.java:146)
       at android.animation.AnimationHandler.access$100(AnimationHandler.java:37)
       at android.animation.AnimationHandler$1.doFrame(AnimationHandler.java:54)
       at android.view.Choreographer$CallbackRecord.run(Choreographer.java:1008)
       at android.view.Choreographer.doCallbacks(Choreographer.java:809)
       at android.view.Choreographer.doFrame(Choreographer.java:740)
       at android.view.Choreographer$FrameDisplayEventReceiver.run(Choreographer.java:995)
       at android.os.Handler.handleCallback(Handler.java:938)
[...]

PR #516, opened 2022-02-16
(@next) fix(android): children views were not recycled

yepMad:
## Summary
Fixes https://github.com/callstack/react-native-pager-view/issues/440

 - This PR overrides the removeAllViews method to ensure that all PagerViewView children are deleted and will not be recycled
 - The removeAllViews function exists in the current version of the project, but does not exist in @next. The motivation was to re-implement it and check if it would work.

You will need **react-navigation@v6.x**

 1. Insert a PagerView
 2. Insert a button with navigation.navigate(...args) or navigation.replace(...args)
	- In the first case give a "goBack" and normally the application would crash.
	 - In the second case you would face a crash at first.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

 - [ ] I added the documentation in  `README.md`


yepMad:
This is very urgent for me, while this PR is under review I sent a package to NPM. I will delete it when this PR is approved.

https://www.npmjs.com/package/@heymad/react-native-pager-view

chakrihacker:
@yepMad I was not able to crash at least.

Deployed a fix in my app. I will observe for a couple of days on crashlytics

Thanks for the fix 🎉

chakrihacker:
One thing to note, if anyone is trying to use this as patch

You also need to patch libraries that use pager-view like material-top-tabs or bottom tabs with imports

just change the default imports to named imports. everything will work

yepMad:
I had an error when I was in development mode, it usually happened through hot-reload. Reviewing my solution, I realized that it didn't make much sense, although it did work. I believe this new one is more in agreement.

@chakrihacker, maybe you want to update your version. I don't think it's critical, I just played it through hot-reload.

troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v6.0.0-rc.1

PR #510, opened 2022-02-08
fix(ios): Crash when rapidly switching pages

doug-lessen:
# Summary

This PR resolves issue: https://github.com/callstack/react-native-pager-view/issues/458

Working demo of no crash when tab switching quickly:

https://user-images.githubusercontent.com/85892917/152933713-fc860143-46b6-4b91-9600-4441fc3171bf.mp4

### What's required for testing (prerequisites)?

Use the example app.

### What are the steps to reproduce (after prerequisites)?

Switch back and forth between tabs super quickly and observe a crash.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [✅] I have tested this on a device and a simulator

PR #518, opened 2022-02-20
Return early when new view controller is the same as the old one

danilobuerger:
In case of setting UIPageViewController.viewControllers to exactly the same controllers, the completion block is not called.
With the introduction of animating in #510, animating will then not be set to NO anymore. 
This will lead to being unable to switch the pages.

# Summary

Fixes a regression in #510 
cc @doug-lessen

## Test Plan

Goto the same page without changing the view, after that switching the page is no longer possible.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -178,6 +178,12 @@
     if (self.reactPageViewController == nil) {
         return;
     }
+
+    NSArray *currentVCs = self.reactPageViewController.viewControllers;
+    if (currentVCs.count == 1 && [currentVCs.firstObject isEqual:controller]) {
+        return;
+    }
+
     __weak ReactNativePageView *weakSelf = self;
     uint16_t coalescingKey = _coalescingKey++;
     


troZee (commented):
Could you provide a example code to reproduce this issue ?

danilobuerger:
@troZee  There is already a working reproduction in the example directory. See the "OnPageScroll Example". You will notice that you can tap through the pages, but once you tap on the page that is already selected, the controls will no longer work.

troZee:
@danilobuerger 
Could you check this fix https://github.com/callstack/react-native-pager-view/releases/tag/v5.4.12

danilobuerger:
@troZee That seems to fix it too but makes no sense to me. Why go through the whole thing, calling `setViewControllers`, waiting for the callback, etc., when we could just return early? My recommendation would be to revert https://github.com/callstack/react-native-pager-view/commit/27e80030c90523664681bb18d0b576c6ab9a4faa and pick this instead.

troZee:
> @troZee That seems to fix it too but makes no sense to me. Why go through the whole thing, calling `setViewControllers`, waiting for the callback, etc., when we could just return early? My recommendation would be to revert [27e8003](https://github.com/callstack/react-native-pager-view/commit/27e80030c90523664681bb18d0b576c6ab9a4faa) and pick this instead.

@danilobuerger thank you for the reviewing and helping me. I did what you have said, since it make a sense at all. Sorry for the inconvenience. I am glad, our cooperation improved this library 🎉 🔥

troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v5.4.14

danilobuerger:
@troZee Awesome! Sorry for not providing any proper testplans initially. I always see things through the contributor lense, not the maintainer lense. I realised that coming back to the PRs a month later I didn't know either what was going on. So in the future I will try to provide test plans straight away!

fobin:
Still crashes for me with 5.4.14 if I rapidly change pages in simulator.

`CoreSimulator 783.5 - Device: iPhone 13 mini (4EC6AC0B-C79C-4438-9F95-428412885F5B) - Runtime: iOS 15.2 (19C51) - DeviceType: iPhone 13 mini
terminating with uncaught exception of type NSException
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Unexpected view controller: <UIViewController: 0x7ff370f902e0>'`

troZee:
@fobin could you try this fix https://github.com/callstack/react-native-pager-view/pull/530 ?

fobin:
At least 5.4.15 also crashes. I guess that has it. 

`Application Specific Information:
dyld4 config: DYLD_ROOT_PATH=/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS.simruntime/Contents/Resources/RuntimeRoot
CoreSimulator 783.5 - Device: iPhone 13 mini (4EC6AC0B-C79C-4438-9F95-428412885F5B) - Runtime: iOS 15.2 (19C51) - DeviceType: iPhone 13 mini
terminating with uncaught exception of type NSException
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Unexpected view controller: <UIViewController: 0x7fd58b7bacd0>'
dyld4 config: DYLD_ROOT_PATH=/Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Library/Developer/CoreSimulator/Profiles/Runtimes/iOS.simruntime/Contents/Resources/RuntimeRoot
abort() called
`

danilobuerger:
@fobin How is this related to this specific PR?

fobin:
I thought this was about fixing the #510 regression so I thought the regression was about #510 not fixing the crash.

danilobuerger:
@fobin this PR doesn't attempt to fix a crash caused by #510. See the summary above.

fobin:
True. My bad. Sorry.

andreialecu:
The change in this PR seems to cause a regression. See https://github.com/callstack/react-native-pager-view/issues/523#issuecomment-1129731887

andreialecu:
Also seems to cause: https://github.com/callstack/react-native-pager-view/issues/543

danilobuerger:
@andreialecu my guess is its unlikely that its the underlying problem to the issues you mentioned as the PR just prevents calling `setViewControllers:direction:animated:completion:` with the same view controller as before which prevents the completion block from being called.

andreialecu:
@danilobuerger well, I have commented out that block of code and the problem gets fixed - I'm not sure what else it breaks though. 

This comment also mentions exactly the same thing: https://github.com/callstack/react-native-pager-view/issues/543#issuecomment-1093455024

danilobuerger:
@andreialecu sure, but then other things break if the completion block is not called. My guess is that a proper fix involves something else. If you provide me with a full repository (not just a code snippet) where your bugs are reproducible, I will take a look.

andreialecu:
There is a repro in https://github.com/callstack/react-native-pager-view/issues/543#issue-1197846706 at the top. I suspect it's the same issue as the one causing #523 

Let me know if you find a solution for #543 and I can test on the issue in #523.

andreialecu:
Full repository repro with the code from #543 at:

https://github.com/andreialecu/rnpv-repro-issue-543

Run `yarn start`, open an iOS simulator (or device) and rotate to landscape mode.

andreialecu:
It appears that changing the return to this:

```objc
    if (currentVCs.count == 1 && [currentVCs.firstObject isEqual:controller]) {
        [self.reactPageViewController setViewControllers:@[controller]
                                           direction:direction
                                            animated:NO
                                          completion:^(BOOL finished) { }];

        return;
    }
```

Resolves the issue, I'm not sure if there's a more elegant way of doing it (no obj-c experience). Or why the call is even needed. 

@danilobuerger perhaps you might have any ideas?

danilobuerger:
Interesting. You can probably pass `nil` for the completion block. Other than that I think if it solves your problems, go for it and create a PR. It shouldn't affect the bug I fixed.

andreialecu:
Opened https://github.com/callstack/react-native-pager-view/pull/565

andreialecu:
#565 was merged, but unfortunately the fix creates another weird issue with the soft keyboard. 

I opened https://github.com/callstack/react-native-pager-view/issues/566 with a video and repro.

PR #520, opened 2022-02-21
fix(ios): Updated index before animation

BrettThePark:
# Summary
On iOS:

If setPage is called and the containing view changes size before animation finishes, the updateDataSource method will be called resetting the page to the preview page.

## Test Plan

I am unable to provide screenshots as the application is unreleased.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [ x] I have tested this on a device and a simulator
- [x ] I added the documentation in `README.md`
- [x ] I updated the typed files (TS and Flow)

### DIFF
--- a/ios/ReactNativePageView.h
+++ b/ios/ReactNativePageView.h
@@ -12,6 +12,7 @@ NS_ASSUME_NONNULL_BEGIN
 @property(nonatomic) NSInteger initialPage;
 @property(nonatomic) NSInteger lastReportedIndex;
 @property(nonatomic) NSInteger currentIndex;
+@property(nonatomic) NSInteger destinationIndex;
 @property(nonatomic) NSInteger pageMargin;
 @property(nonatomic, readonly) BOOL scrollEnabled;
 @property(nonatomic, readonly) BOOL showPageIndicator;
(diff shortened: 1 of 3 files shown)


troZee (commented):
Could you provide a code snippet ?

yepMad:
Hey. This is important. Happens to me when using LazyPagerView. The same problem does not occur on Android. Code example:

`
<PagerContainer
        ref={listViewerRef}
        initialPage={page}
        data={mapping?.questions || []}
        buffer={1}
        maxRenderWindow={5}
        scrollEnabled={false}
        orientation="horizontal"
        transitionStyle="scroll"
        renderItem={renderItem}
        keyExtractor={item => item .id}
        onPageSelected={e => setPage(e.nativeEvent.position)}
      />
`

An useCallback call setPage:
`listViewerRef.current?.setPage(page + 1)`

BrettThePark:
Not sure what the code snippet should contain. This is an iOS only problem. The bug flow is as follows:

The user changes the page to page n+1 with animation. The animation starts, but the index is not set until the animation completes. Before animation completes, the PagerView size is changed, causing the view to be redrawn, but with a page index of n. The end result is that page n is displayed instead of n+1.

yepMad:
I have the same problem. Same flow. The code snippet should be just take the ref and use .setPage(page + 1). I have this problem in the @next version.

PR #493, opened 2021-12-22
Prevent 1 frame flicker on android if initialPage != 0

danilobuerger:
# Summary

Whenever using initialPage != 0 on Android there is a 1 frame flicker where you can see the page at index 0 displayed before the right page at initialPage is displayed. This happens because the setting of the initial page is delayed in https://github.com/callstack/react-native-pager-view/blob/master/android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt#L151
To compensate for not knowing (due to a limitation in rn) how many children are initially added, this also tracks that initial page is only set in the initial render.

### What are the steps to reproduce (after prerequisites)?

Use a PagerView with initialPage != 0 and record the render. Step through the video frame by frame, you will see it jumping on first render.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| Android |    ✅    |

## Checklist

- [x] I have tested this on a device and a simulator

PR #521, opened 2022-02-21
fix(android): cannot set initialPage due to ViewPager2 bug

kyunkakata:
# Summary

On ViewPager2 Android Library has a bug that cannot setCurrentItem at the time function setInitialPage was called (https://stackoverflow.com/questions/56311862/viewpager2-default-position). So, I tried to call setInitialPage when the view was attached to the window to make sure that call setCurrentItem will work normally.
What did I try?
 - Using view.post {} (default) --> sometime it's work, sometime it's not. And when I check isAttachedToWindow sometime return false.
 - Using view.doOnAttach {} --> same like view.post, even when isAttachedToWindow return true.
 - Using view.doOnLayout {} --> works as expected. No need to use hacky solution.
 - Using view.postDelayed {} with 200ms --> Worked. But it's a hacky solution and it doesn't reliable due. And UX is not good ( it was focused on the first page, then focus on the page that was set in initialPage props.
## Test Plan

- set intialPage props = 1 in [example](https://github.com/callstack/react-native-pager-view/blob/master/example/src/PaginationDotsExample.tsx)
- use Android emulator and start the app. The bug will randomly occur. (Due to lack of time, I cannot create a template or bug videos)
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -47,6 +47,10 @@ android {
     sourceCompatibility JavaVersion.VERSION_1_8
     targetCompatibility JavaVersion.VERSION_1_8
   }
+
+  kotlinOptions {
+    jvmTarget = JavaVersion.VERSION_1_8.toString()
+  }
 }
 
 repositories {
@@ -127,4 +131,5 @@ dependencies {
   api 'com.facebook.react:react-native:+'
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   implementation 'androidx.viewpager2:viewpager2:1.0.0'
+  implementation "androidx.core:core-ktx:1.6.0"
 }
--- a/android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
+++ b/android/src/main/java/com/reactnativepagerview/PagerViewViewManager.kt
@@ -2,6 +2,7 @@ package com.reactnativepagerview
 
 import android.view.View
 import android.view.ViewGroup
+import androidx.core.view.doOnLayout
 import androidx.viewpager2.widget.ViewPager2
 import androidx.viewpager2.widget.ViewPager2.OnPageChangeCallback
 import com.facebook.infer.annotation.Assertions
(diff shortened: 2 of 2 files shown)


android/build.gradle
```diff
@@ -47,6 +47,10 @@ android {
     sourceCompatibility JavaVersion.VERSION_1_8
     targetCompatibility JavaVersion.VERSION_1_8
   }
+
+  kotlinOptions {
```
troZee, line 51: Could you explain, why jvm target has been changed ?

android/build.gradle
```diff
@@ -127,4 +131,5 @@ dependencies {
   api 'com.facebook.react:react-native:+'
   implementation "org.jetbrains.kotlin:kotlin-stdlib:$kotlin_version"
   implementation 'androidx.viewpager2:viewpager2:1.0.0'
+  implementation "androidx.core:core-ktx:1.6.0"
```
troZee, line 134: Could you explain, why core-ktx has been added ?

troZee (changes requested):

troZee (changes requested):
@kyunkakata 
Thank you for that fix. I will merge it, once those 2 changes will be removed

android/build.gradle
kyunkakata: Because I used view.doOnLayout {} and it is only available when using core-ktx extension. You can see further information in this doc below https://developer.android.com/reference/kotlin/androidx/core/view/package-summary#(android.view.View).doOnLayout(kotlin.Function1)

android/build.gradle
kyunkakata: When I implemented  "androidx.core:core-ktx:1.6.0", it had a build error. So, I searched and found a solution in StackOverflow. I can't remember exactly what it is. Sorry.

troZee:
https://user-images.githubusercontent.com/12766071/158147281-d41bcd48-15e9-42c0-a3f2-157eeb13b05c.mov

@kyunkakata it does not work on my end in the example :/

kyunkakata:
> Screen.Recording.2022-03-14.at.10.46.59.mov 
> @kyunkakata it does not work on my end in the example :/

Let me check with the example again. Thank you for the review.

troZee:
@kyunkakata 
I noticed https://github.com/callstack/react-native-pager-view/pull/493 here you can find a solution this fix that issue. For now, I am closing this PR. In case of any question, please let me know here in the comments

troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v5.4.13

PR #524, opened 2022-02-25
Do not redisplay currentview if not needed

BrettThePark:
# Summary

iOS First Responder was being lost when React Native was calling goTo on prop change. If goTo is called and the destination page is shown, do not redraw unless it is forced by a layout change.

Have a page with a text field. When the text field is changed, cause the parent to be redrawn with the same page index (causing a goTo to be called). Previously, the first responder would be lost due to the redraw.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

- [ x] I have tested this on a device and a simulator
- [x ] I added the documentation in `README.md`
- [x ] I updated the typed files (TS and Flow)

### DIFF
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -166,7 +166,8 @@
                                  with:initialController
                             direction:UIPageViewControllerNavigationDirectionForward
                              animated:YES
-             shouldCallOnPageSelected:YES];
+             shouldCallOnPageSelected:YES
+                          forceRedraw:YES];
     }
 }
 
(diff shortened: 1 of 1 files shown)


troZee:
https://user-images.githubusercontent.com/12766071/158150032-a862a821-53b0-4fa8-aa62-252667a27d25.mov

It does not work on my end

troZee:
Close since not able to reproduce. Please let me know, if this PR should be reopened for some reason

PR #527, opened 2022-03-02
fix(ios): executing the same setPage does not trigger the goTo method.

alantoa:
# Summary
This is to fix https://github.com/satya164/react-native-tab-view/issues/1322 issue.
When you trigger setPage method more than two times, `view.animating` always False, so the goTo method is not triggered.
I know the last version solved the problem of multiple taps causing crush, but I think this judgment is redundant.

## Test Plan
I tested the Iphone 13, It's work well!

When you trigger setPage method more than two times

See: https://github.com/chshanovskiy/tabview/blob/develop/tabview.gif

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅    |

## Checklist

### DIFF
--- a/ios/ReactViewPagerManager.m
+++ b/ios/ReactViewPagerManager.m
@@ -31,9 +31,7 @@ RCT_EXPORT_VIEW_PROPERTY(layoutDirection, NSString)
             RCTLogError(@"Cannot find ReactNativePageView with tag #%@", reactTag);
             return;
         }
-        if (!animated || !view.animating) {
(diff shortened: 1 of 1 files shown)


jixt:
This fix brings back the crash (when doing rapid switching) that was reported before. I have created another solution that makes sure that both issues are not happening.

https://github.com/callstack/react-native-pager-view/pull/529

alantoa:
@jixt Hi!
I just tested it and it's true.
You are right!

Issue #515, opened 2022-02-16
React Navigation Material Top Tabs stop responding when pressed rapidly.


elmcapp:
## Environment

"@react-navigation/material-top-tabs": "^6.1.1",
"react-native-pager-view": v5.4.11,
"react": "17.0.2",
"react-native": "0.67.2"

## Description

I am using React Navigation Material Top Tabs which depends on this library. 

1.) When I press on tabs rapidly eventually the tabs stop responding. You can swipe left to right on the screen to change pages however pressing on a tab has no response.

2.) This problem also seem to happen if you click on the same tab twice. In this example if you click on CAM then wait 2 seconds and click CAM again then after all tabs become unresponsive. 

## Reproducible Demo

https://user-images.githubusercontent.com/64275658/154208477-c0ad20b4-f1d7-477c-82c1-21b600ddaf43.mp4

dngconsulting:
Same issue here ...

elmcapp:
I notice this even happen if you press slow and if you press a tab and press the same tab again then after no response

coolguy001tv:
> Same issue here with 5.4.11 and react navigation 6 ...
> 
>  Screen.Recording.2022-02-17.at.08.36.33.mov

I temp solve this problem by using https://github.com/callstack/react-native-pager-view/pull/492 with 5.4.9
U may try.

TomCorvus:
Hi there,
same problem here with `5.4.11`. No app crash but the bar doesn't responding.
It's not a fatal crash but it's the same thing on user side in the end.
The problem was mention 5 months ago, any chance to finally eradicate this problem?

Thanks

dngconsulting:
We are trying to ship a product that is blocked waiting for this fix. It's Open Source, that's the game ... Thanks for all you can do regarding this (nasty) bug.

TimMun:
https://github.com/callstack/react-native-pager-view/pull/527

mehthabt:
Thanks @TimMun , this fixed for me

troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v5.4.12

PR #529, opened 2022-03-07
Fixed potential blocking of the navigation

jixt:
# Summary

This PR is a better fix for the following problem. https://github.com/callstack/react-native-pager-view/issues/515. 

The rapid tap causes the setViewControllers function in UIPageViewController to not execute the completion block. When the completion block is not executed, the pager view gets in an invalid state and blocks any attempts to set a different page.

In this fix there is an extra check wether the controller that is set, is the same that is already in the UIPageViewController. If so, no animation is done and the completion block is executed. This makes sure that the completion block is always executed.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅     |

## Checklist

- [X ] I have tested this on a device and a simulator

### DIFF
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -184,10 +184,17 @@
     if (animated == YES) {
         self.animating = YES;
     }
+
+    //
+    // If 'controller' is the current visible controller and the 'animated' flag is set to YES,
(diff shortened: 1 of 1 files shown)


troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v5.4.12

Issue #507, opened 2022-01-25
NSInternalInconsistencyException crashes app


darajava:
## Environment

iOS

## Description

I have some fairly simple code that populates pages and then navigates to one of them. For some reason, if I am generating more than 20/30 pages on iOS, the app crashes with:

*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Unexpected view controller: <UIViewController: 0x7fe24088d100>'

## Reproducible Demo

The code of the component in question:

```
type Props = {
  // Should always be in order
  dates: SliderDate[]
  onDateSelect: (date: Date) => void
}

function sliceIntoChunks(arr: any[], chunkSize: number) {
  const res = []
  for (let i = 0; i < arr.length; i += chunkSize) {
    const chunk = arr.slice(i, i + chunkSize)
    res.push(chunk)
  }

  return res
}

export const DateSlider = (props: Props) => {
  const [weeks, setWeeks] = useState<SliderDate[][]>([])

  const getWeeks = (dates: SliderDate[]): SliderDate[][] => {
    const weekResult: SliderDate[] = []

    /// Fill with blank dates
    const dateTicker = new Date()
    dateTicker.setHours(0, 0, 0, 0)
    dateTicker.setDate(dateTicker.getDate() - 100) // <--- breaks it
    // dateTicker.setDate(dateTicker.getDate() - 10) // <- this one is ok

    // find closest monday
    while (dateTicker.getDay() !== 1) {
      dateTicker.setDate(dateTicker.getDate() + 1)
    }

    console.log("got here")
    for (let i = -10; i < 30; i++) {
      // console.log(dateTicker.getDay())
      if (dateTicker.getDay() !== 0 && dateTicker.getDay() !== 6) {
        weekResult.push({
          date: new Date(dateTicker),
          selected: false,
          recurring: false,
          ordered: false,
          disabled: true,
        })
      }

      dateTicker.setDate(dateTicker.getDate() + 1)
    }

    while (weekResult.length % 5 !== 0) {
      console.log("got here2", weekResult.length)

      if (dateTicker.getDay() !== 0 && dateTicker.getDay() !== 6) {
        weekResult.push({
          date: new Date(dateTicker),
          selected: false,
          recurring: false,
          ordered: false,
          disabled: true,
        })
      }

      dateTicker.setDate(dateTicker.getDate() + 1)
    }

    /// Overwrite with dates from props
    for (const date of props.dates) {
[...]

PR #530, opened 2022-03-11
fix crash with `NSInternalInconsistencyException` on iOS

RishatShamsutdinov:
# Summary

Fix crash with `NSInternalInconsistencyException` on iOS

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -44,7 +44,7 @@
         _dismissKeyboard = UIScrollViewKeyboardDismissModeNone;
         _coalescingKey = 0;
         _eventDispatcher = eventDispatcher;
-        _cachedControllers = [NSHashTable weakObjectsHashTable];
+        _cachedControllers = [NSHashTable hashTableWithOptions:NSHashTableStrongMemory];
         _overdrag = NO;
         _layoutDirection = @"ltr";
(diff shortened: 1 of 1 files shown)


RishatShamsutdinov:
More info about the issue: https://github.com/callstack/react-native-pager-view/issues/507

troZee (changes requested):
@RishatShamsutdinov this solution will create a memory leak, since the reference to the object is strong. 
More explanation here: https://krakendev.io/blog/weak-and-unowned-references-in-swift

RishatShamsutdinov:
The controllers will be relased with the view, it's not memory leak. Just override dealloc method in the view, add breakpoint and check it.

troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v5.4.15

ivan-mumtak:
> https://github.com/callstack/react-native-pager-view/releases/tag/v5.4.15
This bug is still present. I don't have reproducible demo because I'm using react-native-tab-view based on this repo and I have updated pager-view to 5.4.15 but this problem was fixed only partially and everything works fine just 1/3 of time

PR #560, opened 2022-05-09
Only remove child view if it exists

LeviWilliams:
# Summary

This PR simply only removes a child view if its index exists in the view list.

In our case, a page was unmounting that contained a list of reanimated views that had layout animations on them. It appears Reanimated was removing them first causing the pager view to crash when it attempted to remove a child that no longer existed. Removing the layout animations also resolved the issue but that is less than ideal.

Use a pagerview that contains children reanimated views that have an entering or exit animation prop. Unmount the root page they are in and this crash will occur:

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist


troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v5.4.22

Issue #161, opened 2020-04-14
[iOS] delaysContentTouches breaks when child component is Touchable


mrnickel:
# Bug

With the addition of `((UIScrollView *)subview).delaysContentTouches = NO;` on commit https://github.com/react-native-community/react-native-viewpager/commit/8249e31cf04d9961b5fe6ba5dad4ccb443bba48d scrolling through multiple pages of `Touchable`'s only works the first time.

i.e. I have a list of 3 pages, I can scroll once, but subsequent (left or right) via dragging do not work. I _can_ change the page via `setPage` but that's the only way.

## Environment info

`react-native info` output:

```bash
info Fetching system and libraries information...
System:
    OS: macOS 10.15.4
    CPU: (12) x64 Intel(R) Core(TM) i9-8950HK CPU @ 2.90GHz
    Memory: 1.66 GB / 32.00 GB
    Shell: 5.7.1 - /bin/zsh
  Binaries:
    Node: 13.6.0 - /usr/local/bin/node
    Yarn: 1.19.2 - /usr/local/bin/yarn
    npm: 6.13.4 - /usr/local/bin/npm
    Watchman: 4.9.0 - /usr/local/bin/watchman
  SDKs:
    iOS SDK:
      Platforms: iOS 13.4, DriverKit 19.0, macOS 10.15, tvOS 13.4, watchOS 6.2
    Android SDK:
      API Levels: 23, 25, 26, 27, 28, 29
      Build Tools: 25.0.2, 26.0.3, 27.0.3, 28.0.3, 29.0.2
      System Images: android-26 | Google APIs Intel x86 Atom_64, android-27 | Intel x86 Atom_64, android-28 | Google APIs Intel x86 Atom
  IDEs:
    Android Studio: 3.5 AI-191.8026.42.35.6010548
    Xcode: 11.4/11E146 - /usr/bin/xcodebuild
  npmPackages:
    @react-native-community/cli: ^2.0.0 => 2.8.3 
    react: 16.9.0 => 16.9.0 
    react-native: 0.61.5 => 0.61.5 
  npmGlobalPackages:
    react-native-cli: 1.3.0
    react-native-git-upgrade: 0.2.7

```

Library version: 3.2.0 & 3.3.0

## Steps To Reproduce

1. Create a `<ViewPager>` component with 3 `<TouchableWithoutFeedback>` children.
2.  Scroll once to the left.
3. Attempt to scroll either direction.
...

Describe what you expected to happen:

[...]

PR #563, opened 2022-05-12
fix: set delaysContentTouches to YES (fixes iOS touchable issues)

### LINKED ISSUE #554: [bug][ios] Swiping not possible with Touchables from RNGH
## Environment
RN 0.68.1 - current version of rnpv
Tested with bare and Expo.
react-native-gesture-handler: 2.4.2 (also tested with 2.2)

iOS only

## Description
When you use a Flatlist you will most likely import Touchables from react-native-gesture-handler, because they add native buttons and also help to prevent weird activating touchables when you attempt to scroll. I can't use them with RNPV because it breaks swiping. Guess there is some racing with the gestures going on.

## Reproducible Demo
https://snack.expo.dev/@hirbod/pager-view---rngh-touchables-issue

You can swipe the green area, but not the yellow. Only the last yellow box works, since it does use TouchableHighlight by react-native.

hirbod:
# Summary

Set delaysContentTouches to YES (iOS Default)
If the value of this property is true, the scroll view delays handling the touch-down gesture until it can determine if scrolling is the intent. If the value is false , the scroll view immediately calls [touchesShouldBegin(_:with:in:)](https://developer.apple.com/documentation/uikit/uiscrollview/1619418-touchesshouldbegin). The default value is true.

Not sure why this was set to NO in first place, most likely to be able to determine the users touch intent faster, but it breaks native touchables, like those from react-native-gesture-handler.

Fixes #554

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -103,7 +103,7 @@
         if([subview isKindOfClass:UIScrollView.class]){
             ((UIScrollView *)subview).delegate = self;
             ((UIScrollView *)subview).keyboardDismissMode = _dismissKeyboard;
-            ((UIScrollView *)subview).delaysContentTouches = NO;
+            ((UIScrollView *)subview).delaysContentTouches = YES;
             self.scrollView = (UIScrollView *)subview;
         }
     }


hirbod:
There was an issue about this once, which was closed without further explanation: https://github.com/callstack/react-native-pager-view/issues/161

Please describe why this PR can't be accepted, if there is any reason

hirbod:
Another attempt to fix this could also be to SubClass UIScrollView and only to implement

```
- (BOOL)touchesShouldCancelInContentView:(UIView *)view {
    return YES;
}
```

Than we should be able to set `delaysContentTouches` to NO

troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v5.4.16

Issue #490, opened 2021-12-10
PagerView.js:  Attempted import error: 'react-native' does not contain a default export (imported as 'ReactNative').


alphazhe:
## Environment
React native
React native web

## Description

Getting error when run on web
Attempted import error: 'react-native' does not contain a default export (imported as 'ReactNative').

## Reproducible Demo

Clone https://github.com/alphazhe/AntDesignTest/tree/master
`
git checkout master
yarn
yarn web
`
Perform above steps to run above project

Please help me solve this issue.

mohammadgharouni:
I have the same problem on web

PR #579, opened 2022-06-20
fix: #490 react-native-web has no default export

mlecoq:
# Summary

I encounter the issue described in #490. The modification does not change anything on iOS and Android and avoid this issue on react-native-web

You can check it through a newly created expo app

Create an new expo app
Add react-native-pager-view (needed for https://github.com/software-mansion-labs/react-native-url-router in my case-
Start with `yarn web`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist


mlecoq:
There is a second issue on `requireNativeComponent` which is not defined in web case

PR #534, opened 2022-03-16
chore: fabric migration

troZee:
# Current State 
Only Basic iOS features have been implemented so far

# Summary

Add a new arch to the pager library. 

Used docs/code examples: 
https://github.dev/software-mansion/react-native-screens/tree/main/FabricExample
https://github.dev/th3rdwave/react-native-safe-area-context
https://github.dev/cortinico/RNNewArchitectureApp/tree/ios/20220309

New API:
https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/React/Fabric/Mounting/RCTComponentViewProtocol.h

Discussion: 
https://github.com/reactwg/react-native-new-architecture/discussions/16

## Test Plan

2. on iOS
   - Install pods using `RCT_NEW_ARCH_ENABLED=1 pod install` – this is the same command you run to prepare a Fabric build but you also need to run it after a new native library gets added.
3. on Android
   - There are no additional steps required so long you app is configured to build with Fabric – this is typically configured by setting `newArchEnabled=true` in `gradle.properties` file in your project.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ✅    |

## Checklist

- [ ] iOS: migrate all features to the new arch
- [ ] Android: migrate all features to the new arch
- [ ] Create a legacy fallback 
- [ ] Create a separate example for Fabric and Paper arch 

## TO IMPLEMENT ON iOS
- [ ] LTR
- [ ] overdrag
- [ ] issue with jumping tabs

PR #607, opened 2022-09-20
feat(ios): add fabric support on iOS

krozniata:
# Summary

This PR adds **Fabric** support to PagerView, making it compatible with new architecture. 
It adds second implementation ensuring that library will work on both, old and new architecture. 

Supersedes and closes #534

## Test Plan
1. run ``yarn bootstrap-fabric``
2. inside  ``fabricexample`` folder run ``yarn ios``


fabricexample/src/App.tsx
```diff
@@ -115,6 +114,8 @@ export default class App extends React.Component<{}, State> {
           onPageSelected={this.onPageSelected}
           onPageScrollStateChanged={this.onPageScrollStateChanged}
           pageMargin={10}
+          layoutDirection="ltr"
+          overdrag={false}
```
troZee, line 118: We should make it as false inside library

ios/Fabric/RNCPagerViewComponentView.mm
```diff
@@ -0,0 +1,366 @@
...
+
+
+- (void)layoutSubviews {
+    [super layoutSubviews];
+    //Workaround to fix incorrect frame issue
+    [self goTo:_currentIndex animated:NO];
+}
+
+-(void)prepareForRecycle{
+    [super prepareForRecycle];
```
troZee, line 77: Do you know, when this method is called ?

ios/Fabric/RNCPagerViewComponentView.mm
```diff
@@ -0,0 +1,366 @@
...
+    //removes previous page view controller and initializes new one
+    [_nativePageViewController.view removeFromSuperview];
+    _nativePageViewController = nil;
+    
+    [self initializeNativePageViewController];
```
troZee, line 82: Why we have to create a new VC, when view is recycled ?

ios/Fabric/RNCPagerViewComponentView.mm
troZee: https://github.com/facebook/react-native/blob/f3db6cc52792e3006a16408df4ae40f3aee19a86/React/Fabric/Mounting/RCTComponentViewProtocol.h#L112

ios/Fabric/RNCPagerViewComponentView.mm
```diff
@@ -0,0 +1,366 @@
...
+                        direction:direction
+                         animated:animated
+         shouldCallOnPageSelected:YES];
+    
+}
+
+- (void)setPagerViewControllers:(NSInteger)index
+                      direction:(UIPageViewControllerNavigationDirection)direction
+                       animated:(BOOL)animated
+       shouldCallOnPageSelected:(BOOL)shouldCallOnPageSelected {
```
troZee, line 189: I think, we can remove that parameter and document, that progress should be interpolated by reanimated. It would be nice to provide this behaviour in an example

troZee (changes requested):

troZee:
@krozniata could you sync your branch ?

fabricexample/src/App.tsx
krozniata: It's false by default. I can remove it from here as it makes no effect

troZee (approved):

PR #608, opened 2022-09-21
feat(android): add fabric support on Android

okwasniewski:
# Summary

This PR introduces fabric support on Android with backwards compatibility. 

- Configure gradle files for Kotlin
- Implement / extend the generated native interfaces
- Ensure backwards compatibility by splitting implementations into two folders.

## Test Plan

Run `FabricExample` using `yarn android`


android/src/legacy/com/reactnativepagerview/PagerViewViewManager.kt
```diff
@@ -0,0 +1,173 @@
...
+    }
+
+    override fun getExportedCustomDirectEventTypeConstants(): MutableMap<String, Map<String, String>> {
+        return MapBuilder.of(
+                PageScrollEvent.EVENT_NAME, MapBuilder.of("registrationName", "onPageScroll"),
+                PageScrollStateChangedEvent.EVENT_NAME, MapBuilder.of("registrationName", "onPageScrollStateChanged"),
+                PageSelectedEvent.EVENT_NAME, MapBuilder.of("registrationName", "onPageSelected"))
+    }
+
+    override fun receiveCommand(root: NestedScrollableHost, commandId: String?, args: ReadableArray?) {
```
krozniata, line 135: I think that ``commandId`` could be changed to ``commandName``. ``commandId`` suggests that function will expect some kind of ID, but name like ``setPage`` is being passed

android/src/main/java/com/reactnativepagerview/Helper.kt
```diff
@@ -0,0 +1,21 @@
...
+import android.content.Context
+import android.content.ContextWrapper
+import android.view.View
+import com.facebook.react.bridge.ReactContext
+
+
+class Helper {
+    companion object {
+        // https://github.com/facebook/react-native/blob/v0.64.2/ReactAndroid/src/main/java/com/facebook/react/uimanager/UIManagerHelper.java#L138
+        fun getReactContext(view: View): ReactContext? {
```
troZee, line 12: Should we consider to import that function from `UIManagerHelper` ?

troZee (changes requested):

krozniata (approved):

android/src/legacy/com/reactnativepagerview/PagerViewViewManager.kt
okwasniewski: Since this is how codegen generates this function after internal discussion we decided to leave it as is.

troZee (approved):

Issue #420, opened 2021-08-02
RTL is not working


trumdongnat:
## Environment

Android 9, Samsung galaxy s8

## Description

The RTL is working well on example project "BasicPagerViewExample.tsx" but it is not working when I install the package from npm.

## Reproducible Demo

I created a sample project
`npx react-native init TestRTL --template react-native-template-typescript`

And add the package
`yarn add react-native-pager-view`

And the UI like that
```
<SafeAreaView style={backgroundStyle}>
  <StatusBar barStyle={isDarkMode ? 'light-content' : 'dark-content'} />
  <PagerView
    style={{width: '100%', height: '100%'}}
    initialPage={0}
    layoutDirection="rtl">
    <View key="1">
      <Text>First page</Text>
    </View>
    <View key="2">
      <Text>Second page</Text>
    </View>
  </PagerView>
</SafeAreaView>

```

kaiyes:
I have the same issue. rtl doesn't work with next branch.

omar-bizreh:
As a work around from the latest version `5.4.11` I set `layoutDirection` to `ltr`
It will work on RTL but animation is from the other side unfortunately

FadiAboMsalam:
I am facing same issue on layoutDirection any thoughts on this. ?

sebasg0:
@trumdongnat https://github.com/callstack/react-native-pager-view/pull/613 this pr fixes this issue, please help me get it merged

okwasniewski:
This has been fixed by @sebasg0 and released in https://github.com/callstack/react-native-pager-view/releases/tag/v6.0.1

ahmad6242:
RTL still not working in android , in IOS its working.
tested on Pixel 6 API 33.

sebasg0:
What’s your issue @ahmad6242 as for me android always worked.


Can you try doing the following in android.

In manifest add this `android:supportsRtl="true" `. In MainApplication.java add `I18nUtil sharedI18nUtilInstance = I18nUtil.getInstance();
     sharedI18nUtilInstance.allowRTL(this, true); `
This 2 lines of code but don’t forget to do this import at the top `import com.facebook.react.modules.i18nmanager.I18nUtil; `. Let me know if it works!

ahmad6242:
@sebasg0 Ok i will try that tonight. Will let you know. Thanks

PR #613, opened 2022-10-03
fix: on tab selection scroll when rtl enabled

sebasg0:
# Summary
When RTL was enabled the forward direction is the other way around. This is why when isForward was true the for loop in line 264 (of updated code) was not running as when forward was true `_currentIndex` is greater than `index` preventing the for loop to happen. This is why when RTL is enabled I make a new variable to check whether it's enabled or not and if it is the isForward should be the opposite as when LTR is enabled.

the issues this pr close are : 
https://github.com/callstack/react-native-pager-view/issues/420

## Test Plan

https://user-images.githubusercontent.com/83782787/193692343-dd503612-305b-4bd9-914b-138a25464335.mov

https://user-images.githubusercontent.com/83782787/193692558-cfcd8f7f-c8c1-4e67-a69b-9bec33ce687d.mov

Run example project and test changes out

Go to any pagerview example when running example dir app and in app.tsx `I18nManager.forceRTL(true)` for having RTL enable, then refresh app and try any example again. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -247,14 +247,20 @@
         return;
     }
     
-    BOOL isForward = (index > self.currentIndex && [self isLtrLayout]) || (index < self.currentIndex && ![self isLtrLayout]);
+    BOOL isRTL = ![self isLtrLayout];
+    
+    BOOL isForward = (index > self.currentIndex && !isRTL) || (index < self.currentIndex && isRTL);
+
+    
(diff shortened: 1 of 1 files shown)


sebasg0:
Hi @troZee please check this out

troZee:
Thank you @sebasg0 for your 1st contribution 🎉 🔥

troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v6.0.1

PR #641, opened 2022-11-09
Fix indicator animation on setPage

### LINKED ISSUE #636: Incorrect pages transition animation when setPage is called and the new page isn't next or previous from current page
## Environment
Platform: iOS
react-native: 0.69.6
react-native-pager-view: 6.0.2

## Description
When I change the page calling the `setPage` method and the new page value is not next or previous from current page it shows an incorrect transition animation.
For example I have 3 pages. I'm in the page 1. And I will change with `setPage` to page 3 the animation that I'm expecting it's the transition from page 1 (current) to 3 but it shows from page 2 to 3.

## Reproducible Demo
https://user-images.githubusercontent.com/19580424/200635670-bf965360-0dda-4bfb-8c21-f9ea2f1ae7f7.mp4

krozniata:
# Summary

This pull request fixes indicator animation issue on _setPage_ method call on **iOS**. 

When calling _setPage_ method to page which was more than 1 index away for current one, indicator wasn't smoothly animating to new index.

Pull request also fixes incorrect offset value issue, which was causing indicator to "jump" a bit on page change.

Also fixes transition animation issue on **Paper** arch, which was causing to show transition animation from wrong page when navigating with _setPage_ method.

Fixes #636 

## Test Plan

1. open paper/fabric example app
2. open `Basic Example`
3. navigate to last page by pressing `Last` on navigation panel
4. indicator should animate correctly

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/Fabric/RNCPagerViewComponentView.mm
+++ b/ios/Fabric/RNCPagerViewComponentView.mm
@@ -60,7 +60,7 @@ using namespace facebook::react;
         _props = defaultProps;
         _nativeChildrenViewControllers = [[NSMutableArray alloc] init];
         _currentIndex = -1;
-        _destinationIndex = 0;
+        _destinationIndex = -1;
         _layoutDirection = @"ltr";
         _overdrag = NO;
     }
@@ -282,11 +282,11 @@ using namespace facebook::react;
     float offset = 0;
     
     if (self.isHorizontal) {
-        if (self.frame.size.width != 0) {
+        if (scrollView.frame.size.width != 0) {
             offset = (point.x - scrollView.frame.size.width)/scrollView.frame.size.width;
         }
     } else {
-        if (self.frame.size.height != 0) {
+        if (scrollView.frame.size.height != 0) {
             offset = (point.y - scrollView.frame.size.height)/scrollView.frame.size.height;
         }
     }
--- a/ios/Fabric/RNCPagerViewComponentView.mm
+++ b/ios/Fabric/RNCPagerViewComponentView.mm
@@ -60,6 +60,7 @@ using namespace facebook::react;
         _props = defaultProps;
         _nativeChildrenViewControllers = [[NSMutableArray alloc] init];
         _currentIndex = -1;
+        _destinationIndex = 0;
         _layoutDirection = @"ltr";
         _overdrag = NO;
     }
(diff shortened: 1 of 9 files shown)


ios/Fabric/RNCPagerViewComponentView.mm
```diff
@@ -60,6 +60,7 @@ - (instancetype)initWithFrame:(CGRect)frame
         _props = defaultProps;
         _nativeChildrenViewControllers = [[NSMutableArray alloc] init];
         _currentIndex = -1;
+        _destinationIndex = 0;
```
troZee, line 63:
```suggestion
        _destinationIndex = -1;
```
`-1` means that int is not initialized yet, assuming range from 0...n

ios/Fabric/RNCPagerViewComponentView.mm
```diff
@@ -280,11 +283,11 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView {
     
     if (self.isHorizontal) {
         if (self.frame.size.width != 0) {
-            offset = (point.x - self.frame.size.width)/self.frame.size.width;
+            offset = (point.x - scrollView.frame.size.width)/scrollView.frame.size.width;
         }
     } else {
         if (self.frame.size.height != 0) {
```
troZee, line 289:
```suggestion
        if (scrollView.frame.size.height != 0) {
```

ios/Fabric/RNCPagerViewComponentView.mm
```diff
@@ -280,11 +283,11 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView {
     
     if (self.isHorizontal) {
         if (self.frame.size.width != 0) {
```
troZee, line 285:
```suggestion
        if (scrollView.frame.size.width != 0) {
```

ios/ReactNativePageView.m
```diff
@@ -396,23 +380,27 @@ - (void)scrollViewDidScroll:(UIScrollView *)scrollView {
     
     if (self.isHorizontal) {
         if (self.frame.size.width != 0) {
-            offset = (point.x - self.frame.size.width)/self.frame.size.width;
```
troZee, line 399: Please apply all above comments to the paper arch

troZee (changes requested):

Bardiamist:
Nice!

PR #651, opened 2022-11-21
fix(iOS): fix setPage issue

### LINKED ISSUE #649: `setPage` can stop working if multiple user interactions happen quickly
## Environment

iOS 16.1, reproduced on iPhone 14 Simulator and iPhone 7.

## Description

If you call the function `setPage` and very quickly after it swipe to another page, the `setPage` can stop working. Further calls to the function will not actually set the page, but are ignored completely. This can happen e.g. with a tab view, where the user presses a tab title and then swipes. The problem does not occur with `setPageWithoutAnimation`.

I think the issue in here: https://github.com/callstack/react-native-pager-view/blob/7ff5c1518b2c56fd8499786a8325aded15b179bb/ios/ReactViewPagerManager.m#L33 and here https://github.com/callstack/react-native-pager-view/blob/7ff5c1518b2c56fd8499786a8325aded15b179bb/ios/ReactNativePageView.m#L179

The swipe or the `setPage` responds with `finished` being `false` which causes the `animating` value to stay `true` blocking further animating `setPage` calls. This would also explain why the `setPageWithoutAnimation` does not have the same bug.

## Reproducible Demo

The issue can be reproduced with the [CustomTabBarExample](https://github.com/callstack/react-native-pager-view/blob/master/example/src/tabView/CustomTabBarExample.tsx) in the project repo. Here is a video of it (it does not happen on the first try, but does on the second): 

https://user-images.githubusercontent.com/33930143/202645438-7d3a44bb-eaa2-4103-9af3-e9490a7b91f9.mp4

krozniata:
# Summary

This pull request fixes issue causing _setPage_ method to stop working. 
It does that by disabling user interactions while **Pager View** is animating to new page.

Fixes #649 
## Test Plan

As described in issue #649 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/Fabric/RNCPagerViewComponentView.mm
+++ b/ios/Fabric/RNCPagerViewComponentView.mm
@@ -176,6 +176,8 @@ using namespace facebook::react;
 - (void)goTo:(NSInteger)index animated:(BOOL)animated {
     NSInteger numberOfPages = _nativeChildrenViewControllers.count;
     
+    _nativePageViewController.view.userInteractionEnabled = NO;
+    
     _destinationIndex = index;
     
     
@@ -210,6 +212,7 @@ using namespace facebook::react;
             int position = (int) index;
             strongEventEmitter.onPageSelected(RNCViewPagerEventEmitter::OnPageSelected{.position =  static_cast<double>(position)});
             strongSelf->_currentIndex = index;
+            strongSelf->_nativePageViewController.view.userInteractionEnabled = YES;
         }
     }];
 }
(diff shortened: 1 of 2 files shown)


troZee (approved):

okwasniewski (approved):

jafar-jabr:
This introduce a very bad bug,
it makes the MaterialTopTabNavigator screen freeze until I navigate to and other screen.

troZee:
@jafar-jabr Could you open an issue and provide an example repro ?

PR #665, opened 2022-12-03
feat: back gesture with full screen swipe

intergalacticspacehighway:
# Summary

- React Native Screen has a prop named [fullScreenGestureEnabled](https://reactnavigation.org/docs/native-stack-navigator/#fullscreengestureenabled) that enables back gesture on full-screen swipe. This PR makes that prop work nicely with pager view by disabling it's scroll when index is 0 and user is swiping back, so the RN Screen gesture can take over.
- Resolves - https://github.com/PedroBern/react-native-collapsible-tab-view/issues/248
- Also updated react navigation and screens dependency because this prop is only available in the newer version 😅 

## Test Plan
- Add `fullScreenGestureEnabled` on react navigation's native stack and try to swipe when the first page is reached in pager view.
 

https://user-images.githubusercontent.com/23293248/205438463-fafa54a2-664e-4ad0-8f82-04985e2cc399.mov

### Test with RTL enabled

`I18nManager.forceRTL(true);`

https://user-images.githubusercontent.com/23293248/205438337-dc0371a8-f871-4581-993f-f1bdfd2fc0de.mov

- Run yarn install and pod install in example
- Add `fullScreenGestureEnabled` in react navigation screen options

- Open any pager view example and `swipe to go back` with `fullScreenGestureEnabled` should work on the first page.
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist


hirbod:
Using this patch for many many months already, can confirm that it just works (TM)

ios/ReactNativePageView.m
```diff
@@ -461,4 +466,24 @@ - (NSString *)determineScrollDirection:(UIScrollView *)scrollView {
 - (BOOL)isLtrLayout {
     return [_layoutDirection isEqualToString:@"ltr"];
 }
+
+
+- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecognizeSimultaneouslyWithGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer {
+    
+    // The below snippet disables the pager view's scrollview's scroll when current index is 0 and user is swiping back. Useful for fullScreenGestureEnabled in react-native-screens
```
satya164, line 473: I wonder if this should also happen if it's the last page and the user is swiping in the other direction. It's not related to native stack but it would make it work when it's nested in something else with horizontal gestures. For example, maybe drawer on the right.

intergalacticspacehighway: yes, my bad. The default return value should be [false](https://developer.apple.com/documentation/uikit/uigesturerecognizerdelegate/1624208-gesturerecognizer?language=objc). It should only recognize simultaneously if the other recognizer is the pan responder that we've attached. Fixed [here](https://github.com/callstack/react-native-pager-view/pull/665/commits/5a517447ada1c55496b4de17a4a7067ff7f1e558). Thanks for catching it!

intergalacticspacehighway: okay. I misunderstood what you meant. I think in that case, it'll default to `self.scrollView.panGestureRecognizer.enabled = self.scrollEnabled`  i.e. default scroll enabled setting (since it's not a back gesture). So I guess it should be fine. I think something like a Drawer would need to prevent the pager gesture too. Happy to test it though.

ios/ReactNativePageView.m
```diff
@@ -461,4 +466,28 @@ - (NSString *)determineScrollDirection:(UIScrollView *)scrollView {
...
+            self.scrollView.panGestureRecognizer.enabled = false;
+            return NO;
+        }
+        
+        self.scrollView.panGestureRecognizer.enabled = self.scrollEnabled;
+        return YES;
+    }
+    
+    // Resetting here to default in case we add more recognizer.
+    self.scrollView.panGestureRecognizer.enabled = self.scrollEnabled;
```
intergalacticspacehighway, line 489: This LOC will be unreachable as of now as we added only one recognizer. But still resetting the value for sanity.

ios/ReactNativePageView.m
intergalacticspacehighway: Did some quick testing with the drawer, this method doesn't run for the drawer as the gesture never reaches the page view (because of Overlay).

https://user-images.githubusercontent.com/23293248/205469645-b670a294-7c44-4a4d-937c-715c2107f994.mov


I still think that even if it runs for some horizontal gesture, it should be "mostly" fine.
### Scenario - Someone adds a gesture recognizer to a parent view of the pager view.
- By default, the parent recognizer won't receive gesture events. However, they would if a user is swiping "back" and they're on page 0 (because we'd disable the scroll). tbh I am not sure if it's a bug or a feature 😅

intergalacticspacehighway: Relatively Safer alternative that works. One downside is that it depends upon the member class name for pan gesture used in RN Screens.

```objc

// The below snippet disables the pager view's scrollview's scroll when current index is 0 and user is swiping back. Useful for fullScreenGestureEnabled in RN Screens
- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecognizeSimultaneouslyWithGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer {

    // Recognize simultaneously only if the other gesture is RN Screen's pan gesture (one that is used to perform fullScreenGestureEnabled)
    if (gestureRecognizer == self.panGestureRecognizer && [NSStringFromClass([otherGestureRecognizer class]) isEqual: @"RNSPanGestureRecognizer"]) {
        UIPanGestureRecognizer* panGestureRecognizer = (UIPanGestureRecognizer*) gestureRecognizer;
        CGPoint velocity = [panGestureRecognizer velocityInView:self];
        BOOL isLTR = [self isLtrLayout];
        BOOL isBackGesture = (isLTR && velocity.x > 0) || (!isLTR && velocity.x < 0);
        
        if (self.currentIndex == 0 && isBackGesture) {
            self.scrollView.panGestureRecognizer.enabled = false;
        } else {
            self.scrollView.panGestureRecognizer.enabled = self.scrollEnabled;
        }
        
        return YES;
    }
    
    self.scrollView.panGestureRecognizer.enabled = self.scrollEnabled;
    return NO;
}

```

example/package.json
```diff
@@ -11,16 +11,16 @@
   },
   "dependencies": {
     "@react-native-community/masked-view": "^0.1.10",
-    "@react-navigation/native": "^5.8.10",
-    "@react-navigation/native-stack": "5.0.4",
-    "@react-navigation/stack": "^5.12.8",
+    "@react-navigation/native": "^6.0.16",
```
troZee, line 14: Can we blindly bump a major version of React Navigation ? Do we have to do a migration ?

troZee (changes requested):
Hey @intergalacticspacehighway 
Thank you for adding a new feature for the Pager 🔥  Could you add the same implementation for a fabric here: https://github.com/callstack/react-native-pager-view/blob/master/ios/Fabric/RNCPagerViewComponentView.mm . Unfortunately, we have 2 implementations (paper and fabric).

ouabing:
Can confirm this PR solves the issue can't swipe back when pager view is inside a react navigation screen

andreialecu:
Any plan to continue work/update this to the current master? There has been a lot of activity in the repo, and I'm hopeful that it might get merged this time.

hirbod:
I think Nishan did not continue here because they started rewriting to ScrollView. @intergalacticspacehighway right ?

intergalacticspacehighway:
Closing this in favor of https://github.com/callstack/react-native-pager-view/pull/705!

PR #672, opened 2022-12-19
feat(iOS): change Fabric implementation to UIScrollView

krozniata:
# Summary

This PR introduces _PagerView_ **Fabric** implementation rewrite from **UIPageView** to **UIScrollView**.

It is based on React Native's **UIScrollView** implementation for  **Fabric**.

## Motivation

As described in #673, we wanted to have more control over native side of _PagerView_ which could be achieved by switching from using **UIPageViewController** to **UIScrollView**. 

This change makes _PagerView_ easier to maintain, because some solutions and workarounds used in previous implementation were adding more complexity and were causing some additional bugs. 

Switching to **UIScrollView** allowed to reduce amount of code and complexity, while maintaining same functionality.
In future it will also be easier to add new features or e.g. custom animations to _PagerView_.

## Test Plan

Every example should work as expected (only Fabric architecture)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/common/cpp/react/renderer/components/RNCViewPager/RNCViewPagerShadowNode.cpp
+++ b/common/cpp/react/renderer/components/RNCViewPager/RNCViewPagerShadowNode.cpp
@@ -1,0 +1,35 @@
+#include "RNCViewPagerShadowNode.h"
+
+#include <react/debug/react_native_assert.h>
+#include <react/renderer/core/LayoutMetrics.h>
+
+namespace facebook {
+namespace react {
+
+const char RNCViewPagerComponentName[] = "RNCViewPager";
+
+void RNCViewPagerShadowNode::updateStateIfNeeded() {
+    ensureUnsealed();
+    
+    auto contentBoundingRect = Rect{};
+    for (const auto &childNode : getLayoutableChildNodes()) {
+        contentBoundingRect.unionInPlace(childNode->getLayoutMetrics().frame);
+    }
+    
+    auto state = getStateData();
+    
+    if (state.contentBoundingRect != contentBoundingRect) {
+        state.contentBoundingRect = contentBoundingRect;
+        setStateData(std::move(state));
+    }
+}
+
+#pragma mark - LayoutableShadowNode
+
+void RNCViewPagerShadowNode::layout(LayoutContext layoutContext) {
+    ConcreteViewShadowNode::layout(layoutContext);
+    updateStateIfNeeded();
+}
+
+}
+}
--- a/common/cpp/react/renderer/components/RNCViewPager/RNCViewPagerShadowNode.h
+++ b/common/cpp/react/renderer/components/RNCViewPager/RNCViewPagerShadowNode.h
@@ -1,0 +1,31 @@
+#pragma once
+
+#include <react/renderer/components/RNCViewPager/EventEmitters.h>
+#include <react/renderer/components/RNCViewPager/Props.h>
+#include <react/renderer/components/RNCViewPager/RNCViewPagerState.h>
+#include <react/renderer/components/view/ConcreteViewShadowNode.h>
+#include <react/renderer/core/LayoutContext.h>
+
+namespace facebook {
+namespace react {
+
+extern const char RNCViewPagerComponentName[];
+
+class RNCViewPagerShadowNode final : public ConcreteViewShadowNode<
+                                                RNCViewPagerComponentName,
+                                                RNCViewPagerProps,
+                                                RNCViewPagerEventEmitter,
+                                                RNCViewPagerState> {
+public:
+    using ConcreteViewShadowNode::ConcreteViewShadowNode;
+    
+#pragma mark - LayoutableShadowNode
+    
+    void layout(LayoutContext layoutContext) override;
+    
+private:
+    void updateStateIfNeeded();
+};
+
+}
+}
--- a/fabricexample/src/NestPagerView.tsx
+++ b/fabricexample/src/NestPagerView.tsx
@@ -20,12 +20,12 @@ export function NestPagerView() {
       >
         <View
           key="1"
-          style={{ backgroundColor: BGCOLOR[0] }}
+          style={[styles.page, { backgroundColor: BGCOLOR[0] }]}
           collapsable={false}
         >
           <LikeCount />
         </View>
-        <View key="2" collapsable={false}>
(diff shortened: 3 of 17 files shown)


common/cpp/react/renderer/components/RNCViewPager/RNCViewPagerShadowNode.cpp
```diff
@@ -0,0 +1,35 @@
+#include "RNCViewPagerShadowNode.h"
```
troZee, line 1:
```suggestion
// https://github.com/facebook/react-native/blob/main/ReactCommon/react/renderer/components/scrollview/ScrollViewShadowNode.cpp
#include "RNCViewPagerShadowNode.h"
```

common/cpp/react/renderer/components/RNCViewPager/RNCViewPagerShadowNode.h
```diff
@@ -0,0 +1,31 @@
+#pragma once
```
troZee, line 1: https://github.com/facebook/react-native/blob/main/ReactCommon/react/renderer/components/scrollview/ScrollViewShadowNode.h

src/PagerView.tsx
```diff
@@ -149,7 +149,13 @@ export class PagerView extends React.Component<PagerViewProps> {
         ref={(ref) => {
           this.pagerView = ref;
         }}
-        style={this.props.style}
+        style={[
+          this.props.style,
+          {
+            flexDirection:
+              this.props.orientation === 'vertical' ? 'column' : 'row',
```
troZee, line 156: Breaking change

src/utils.tsx
```diff
@@ -1,17 +1,16 @@
...
+import { View } from 'react-native';
 
 export const childrenWithOverriddenStyle = (children?: ReactNode) => {
   return Children.map(children, (child) => {
     const element = child as React.ReactElement;
     return (
       // Add a wrapper to ensure layout is calculated correctly
-      <View style={StyleSheet.absoluteFill} collapsable={false}>
-        {/* @ts-ignore */}
+      <View style={{ height: '100%', width: '100%' }} collapsable={false}>
```
troZee, line 9: Probably we should refine this function, since styles should not be overridden

troZee (changes requested):

common/cpp/react/renderer/components/RNCViewPager/RNCViewPagerShadowNode.cpp
thymikee: any way to generate that with codegen or suggest the feature request?

troZee: Right now, the codegen generates those files inside a build folder. We can suggest a feature request and request a comment from other developers, on what they think about it.

src/utils.tsx
```diff
@@ -1,17 +1,26 @@
...
+          paddingHorizontal: pageMargin / 2,
+        }}
+        collapsable={false}
+      >
         {React.cloneElement(element, {
           ...element.props,
           // Override styles so that each page will fill the parent.
-          style: [element.props.style, StyleSheet.absoluteFill],
+          style: [element.props.style, { height: '100%', width: '100%' }],
         })}
```
satya164, line 24: Unrelated to this PR, but I wonder why is this needed given there's already a wrapper view. Changing the dimensions of the view may be unnecessary at best or undesirable at worst.

krozniata: I will take a look at it!

fabricexample/src/NestPagerView.tsx
```diff
@@ -82,7 +82,7 @@ export function NestPagerView() {
         </View>
         <View
           key="3"
-          style={{ backgroundColor: BGCOLOR[3] }}
+          style={[styles.page, { backgroundColor: BGCOLOR[3] }]}
```
troZee, line 85: We have to highlight it as a breaking change

src/utils.tsx
krozniata: Removed style overwriting. Now each child is only wrapped with wrapper view.

src/utils.tsx
```diff
@@ -1,18 +1,21 @@
...
     return (
-      // Add a wrapper to ensure layout is calculated correctly
-      <View style={StyleSheet.absoluteFill} collapsable={false}>
-        {/* @ts-ignore */}
-        {React.cloneElement(element, {
-          ...element.props,
-          // Override styles so that each page will fill the parent.
-          style: [element.props.style, StyleSheet.absoluteFill],
-        })}
+      <View
```
okwasniewski, line 10: I think we need to pass there `child.key` so the outer view has proper key and react can manage it's views efficiently. I remember that's what @satya164 mentioned on one of our internal meetings.

ios/Fabric/RNCPagerViewComponentView.mm
```diff
@@ -260,130 +151,71 @@ - (void)scrollViewWillBeginDragging:(UIScrollView *)scrollView {
...
 }
 
-- (UIViewController *)pageViewController:(UIPageViewController *)pageViewController
-      viewControllerBeforeViewController:(UIViewController *)viewController {
-    UIPageViewControllerNavigationDirection direction = [self isLtrLayout] ? UIPageViewControllerNavigationDirectionReverse : UIPageViewControllerNavigationDirectionForward;
-    return [self nextControllerForController:viewController inDirection:direction];
+- (void)setPageWithoutAnimation:(NSInteger)index {
+    CGPoint targetOffset = [self isHorizontal] ? CGPointMake(_scrollView.frame.size.width * index, 0) : CGPointMake(0, _scrollView.frame.size.height * index);
+    
+    [_scrollView setContentOffset:targetOffset animated:NO];
```
okwasniewski, line 214: This part is duplicated with the `setPage` method. Maybe you could create one methods that's called in those two places.

src/utils.tsx
krozniata: @satya164 `React.cloneElement` has been removed 👌

ios/Fabric/RNCPagerViewComponentView.mm
krozniata: Added `getPageOffset` method

Issue #673, opened 2022-12-20
☂️ Pager View Improvements


troZee:
### Intro

It is been a while since I am maintaining the `react-native-pager-view`. Previously it was part of the RN core, but then it has been extracted into a separate package. It has used 2 versions of ViewPager (v1, v2) and UIPageView. I tried to fix all issues, but some of them required some workarounds that were not stable enough. Those workarounds created other issues, hence other workarounds were used to fix that. To be honest, I was a little bit exhausted from it, hence together with @krozniata  and @okwasniewski  decided to build those components from the scratch 

### Next branch
@alpha0010  did a great job by implementing the `LazyViewPager` component. We released his changes into the `next` branch and we had the plan to merge it into the main branch. But we did some internal tests on real applications and in some cases, it did not work well. Internally we used another workaround to achieve the lazy effect. Moreover, we did not want to add additional complexity, when it did not bring more benefits and would add additional bugs to fix. With new architecture implementing a lazy approach should be much easier. Unfortunately, we won’t merge the next branch into the main. We will implement a lazy approach using a new architecture. If you would like to maintain `LazyViewPager`, feel free to make a fork

### New architecture support

Meta has released a new architecture called fabric. In this library, we would like to focus first on the new architecture. After that, we will evaluate, if it would be worth migrating a new solution to the old arch (Right now it is hard to estimate it based on a priori knowledge) 

### Motivation

We would like to have more control over native components to easily fix issues or add new features. Right now, those components are black boxes, which are hard to maintain. Moreover, `UIPageViewController` is not maintained anymore and is not used in the iOS world. Unfortunately, we cannot use `UICollectionView`, because it does not work well in the RN world. 

The same thing refers `RecyclerView` component for Android. The currently used `ViewPager2` component on Android brought more issues than benefits since it uses the earlier-mentioned `RecyclerView`.

troZee:
Here you can see a draft implementation on an iOS platform: https://github.com/callstack/react-native-pager-view/pull/672

alexeykomov:
@troZee Hello, thanks! What's the complexity with `UICollectionView`? Is it the fact that `reactSubviews` expose subviews of view, but there's no such equivalent of cells for a `UICollectionView`?

Also, side question - is React Native's `FlatList` implemented with plain `UIScrollView`, not `UICollectionView`?

Thanks.

r1me75:
[...]

PR #681, opened 2023-01-18
feat(iOS): rewrite old arch to use UIScrollView

okwasniewski:
# Summary

This PR rewrites old architecture of react-native-pager-view to UIScrollView on iOS. 

## Test Plan

Launch the example app.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

## Known issues
- [ ] Swiping right on last page is stuck on settling when overdrag is disabled.

### DIFF
--- a/ios/RNCPagerView.h
+++ b/ios/RNCPagerView.h
@@ -5,7 +5,7 @@
 
 NS_ASSUME_NONNULL_BEGIN
 
-@interface ReactNativePageView: UIView
+@interface RNCPagerView: UIView
 
 - (instancetype)initWithEventDispatcher:(id<RCTEventDispatcherProtocol> )eventDispatcher;
 
--- a/ios/RNCPagerView.h
+++ b/ios/RNCPagerView.h
@@ -11,8 +11,6 @@ NS_ASSUME_NONNULL_BEGIN
 
 @property(nonatomic) NSInteger initialPage;
 @property(nonatomic) NSString* orientation;
-@property(nonatomic) NSInteger currentIndex;
-@property(nonatomic) NSInteger pageMargin;
 @property(nonatomic, readonly) UIScrollViewKeyboardDismissMode dismissKeyboard;
 @property(nonatomic, copy) RCTDirectEventBlock onPageSelected;
 @property(nonatomic, copy) RCTDirectEventBlock onPageScroll;
--- a/ios/RNCPagerView.m
+++ b/ios/RNCPagerView.m
@@ -58,6 +58,10 @@
 - (void)removeReactSubview:(UIView *)subview {
     [super removeReactSubview:subview];
     [subview removeFromSuperview];
+    
+    if ([self getCurrentPage] >= self.reactSubviews.count - 1) {
+        [self goTo:self.reactSubviews.count - 1 animated:false];
+    }
 }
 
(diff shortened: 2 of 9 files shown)


ios/ReactNativePageView.m
```diff
@@ -127,338 +93,88 @@ - (void)shouldDismissKeyboard:(NSString *)dismissKeyboard {
...
-    } else {
-        [self goTo:newIndex animated:NO];
+    if (self.initialPage != 0) {
+        [self goTo:self.initialPage animated:false];
     }
 }
 
 - (void)disableSwipe {
-    self.reactPageViewController.view.userInteractionEnabled = NO;
+    self.scrollView.userInteractionEnabled = NO;
```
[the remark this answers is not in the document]
okwasniewski: You are right, changed it 👍🏻

ios/RNCPagerView.h
```diff
@@ -5,23 +5,19 @@
...
 
-- (instancetype)initWithEventDispatcher:(RCTEventDispatcher *)eventDispatcher;
+- (instancetype)initWithEventDispatcher:(id<RCTEventDispatcherProtocol> )eventDispatcher;
 
 @property(nonatomic) NSInteger initialPage;
-@property(nonatomic) NSInteger lastReportedIndex;
-@property(nonatomic) NSInteger destinationIndex;
+@property(nonatomic) NSString* orientation;
 @property(nonatomic) NSInteger currentIndex;
 @property(nonatomic) NSInteger pageMargin;
```
krozniata, line 15: I think this can be removed, because it's not used anymore inside `RNCPagerView.m`

ios/RNCPagerView.m
```diff
@@ -0,0 +1,182 @@
...
+#import <math.h>
+
+@interface RNCPagerView () <UIScrollViewDelegate>
+
+@property(nonatomic, strong) id<RCTEventDispatcherProtocol> eventDispatcher;
+
+@property(nonatomic, strong) UIScrollView *scrollView;
+@property(nonatomic, strong) UIView *containerView;
+
+@property(nonatomic, assign) CGPoint lastContentOffset;
```
krozniata, line 20: Why do we need this property?

krozniata (approved):

andreialecu:
I'm curious about how this is different from a regular RN ScrollView with `pagingEnabled: true`?

I had issues with RN ScrollView in the past and moved to `react-native-pager-view` specifically to solve those.

Issues were: it would randomly get stuck between pages at some points, and performance seemed worse in general, especially with Reanimated and complex interactions.

okwasniewski:
Hey @andreialecu!   

There are quite a lot of differences between those two implementations. The Core RN ScrollView is a generic scrolling component written to support as many use cases as possible to be the “go-to” scrolling component. 
 
Both implementations are based on the native `UIScrollView` component, but pager's implementation is focused on providing the best paging experience possible, while keeping the same API as old implementation. You can read more about the motivation here: https://github.com/callstack/react-native-pager-view/issues/673
 
Also, we are going to soon release pager-view `7.0.0-next` it would be awesome If you could test it out on your specific use case and let us know if you have any performance issues.

PR #693, opened 2023-02-13
feat: make pager controlled

krozniata:
# Summary

This PR makes **Pager View** a controlled component, giving users the ability to control the current page with new `page` and `animated` properties.

It also changes all examples to use the controlled version of **Pager View**, as this will be the recommended way of using it.

This feature will only be available from version `7.x.x`

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist


ios/RNCPagerViewManager.m
```diff
@@ -7,6 +7,9 @@ @implementation RNCPagerViewManager
 
 RCT_EXPORT_MODULE(RNCViewPager)
 
+RCT_EXPORT_VIEW_PROPERTY(page, NSInteger)
+RCT_EXPORT_VIEW_PROPERTY(animated, BOOL)
```
troZee, line 11: I have a question about the `onPageSelected` behavior.  Should it behave in the same way as uncontrolled flow? 

Because in controlled flow, this callback is not necessary to implement?

src/PagerViewNativeComponent.ts
```diff
@@ -24,6 +24,8 @@ export type OnPageScrollStateChangedEventData = Readonly<{
 }>;
 
 export interface NativeProps extends ViewProps {
+  page?: Int32;
```
troZee, line 27: Could you extract those props to a separate interface?

ios/RNCPagerViewManager.m
krozniata: I think so. When setting the page with 'page' props, the user will still be able to call some functions at the end of the page change transition.

android/src/fabric/java/com/reactnativepagerview/PagerViewViewManager.kt
```diff
@@ -105,6 +105,17 @@ class PagerViewViewManager : ViewGroupManager<NestedScrollableHost>(), RNCViewPa
         return PagerViewViewManagerImpl.needsCustomLayoutForChildren()
     }
 
+    @ReactProp(name = "page")
+    override fun setPage(view: NestedScrollableHost?, value: Int) {
+        val animated = PagerViewViewManagerImpl.animated
```
okwasniewski, line 110: I would pass the `PagerViewViewManagerImpl.animated` directly to the function

README.md
```diff
@@ -184,8 +186,8 @@ For advanced usage please take a look into our [example project](https://github.
...
 | ------------------------------------------ | :--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------: | :------: |
-| `setPage(index: number)`                   |                                                                      Function to scroll to a specific page in the PagerView. Invalid index is ignored.                                                                       |   both   |
-| `setPageWithoutAnimation(index: number)`   |                                                                      Function to scroll to a specific page in the PagerView. Invalid index is ignored.                                                                       |   both   |
+| `setPage(index: number)`                   |                                                                      Function to scroll to a specific page in the PagerView. Invalid index is ignored. The recommended way is using the `page` and `animated` props                                                         |   both   |
```
okwasniewski, line 189: I think we might need a broader explanation of the controlled vs uncontrolled usage in docs. But this can be added later on before merging 7.0.0 to master

android/src/main/java/com/reactnativepagerview/PagerViewViewManagerImpl.kt
```diff
@@ -7,6 +7,8 @@ import com.facebook.react.uimanager.PixelUtil
 object PagerViewViewManagerImpl {
     const val NAME = "RNCViewPager"
 
+    var animated = true
```
okwasniewski, line 10: Do we need to set this?  

Since the prop is defining it's default value here: `@ReactProp(name = "animated", defaultBoolean = true)`

README.md
krozniata: Yep, I agree! Version `7.0.0` needs some changes in docs, so it can be done later

android/src/fabric/java/com/reactnativepagerview/PagerViewViewManager.kt
krozniata: Passed `PagerViewViewManagerImpl.animated` directly

android/src/main/java/com/reactnativepagerview/PagerViewViewManagerImpl.kt
krozniata: It has to be set, because `lateinit` is supported for properties of object types only, not primitive types

ios/Fabric/RNCPagerViewComponentView.mm
```diff
@@ -68,6 +75,14 @@ - (void)updateProps:(const facebook::react::Props::Shared &)props oldProps:(cons
         [_scrollView setScrollEnabled:newScreenProps.scrollEnabled];
     }
     
+    if (newScreenProps.animated != _animated) {
+        _animated = newScreenProps.animated;
+    }
+    
+    if (newScreenProps.page != _page && newScreenProps.page >= 0 && newScreenProps.page < _containerView.subviews.count) {
```
okwasniewski, line 82: Here we can check reactSubviews count, it would mostly do the same but be a bit more clear what are we checking here. But you can leave it like so.

krozniata: Changed it in `RNVViewPager.m`, because there is no `reactSubiews` in fabric implementation

troZee:
RN this PR is on hold, bc the sync mechanism in case of failure does not work

PR #698, opened 2023-03-16
fix: adjust `canChildScroll` based on `scrollEnabled`

lanpai:
# Summary

Currently when disabling scroll using the `scrollEnabled` prop, the pager still consumes user inputs on Android. This adjusts `NestedScrollableHost` such that that it no longer consumes touch events when `isUserInputEnabled` is disabled on the child pager by the `scrollEnabled` prop.

A set up with two nested pagers aligned in the same direction on an Android device.

On iOS, expected behavior works correctly.

Attempt to scroll a multi-page pager via touch when it's nested inside of another pager on the same axis in a direction where there's an available page. Previously, this would prevent both pagers from scrolling when the parent pager has `scrollEnabled=true` and the child pager has `scrollEnabled=false`. Now, it disables event handling on the child pager allowing the parent pager to scroll.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    N/A     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativepagerview/NestedScrollableHost.kt
+++ b/android/src/main/java/com/reactnativepagerview/NestedScrollableHost.kt
@@ -43,6 +43,10 @@ class NestedScrollableHost : FrameLayout {
   }
 
   private fun canChildScroll(orientation: Int, delta: Float): Boolean {
+    if ((child as? ViewPager2)?.isUserInputEnabled == false) {
+      return false
+    }
+
     val direction = -delta.sign.toInt()
     return when (orientation) {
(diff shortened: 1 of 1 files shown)


troZee:
Hey @lanpai 
Thank you for submitting a PR. Could you provide a simple code snippet, which I will be able to reproduce it?

yuhr:
I've experienced the same problem and workarounded by the exact same change using `patch-package`. This should definitely be merged to match behavior between iOS and Android.

yuhr:
@MrRefactor Why is this closed silently?

PR #705, opened 2023-03-24
feat: back gesture with full screen swipe

intergalacticspacehighway:
Previous PR - https://github.com/callstack/react-native-pager-view/pull/665

# Summary

- React Native Screen has a prop named [fullScreenGestureEnabled](https://reactnavigation.org/docs/native-stack-navigator/#fullscreengestureenabled) that enables back gesture on full-screen swipe. This PR makes that prop work nicely with pager view by disabling it's scroll when index is 0 and user is swiping back, so the RN Screen gesture can take over.
- Resolves - https://github.com/PedroBern/react-native-collapsible-tab-view/issues/248
- Also updated react navigation and screens dependency because this prop is only available in the newer version 😅 

## Test Plan
- Add `fullScreenGestureEnabled` on react navigation's native stack and try to swipe when the first page is reached in pager view.
- Test fabricexample too with the same changes.
 

https://user-images.githubusercontent.com/23293248/205438463-fafa54a2-664e-4ad0-8f82-04985e2cc399.mov

### Test with RTL enabled

`I18nManager.forceRTL(true);`

https://user-images.githubusercontent.com/23293248/205438337-dc0371a8-f871-4581-993f-f1bdfd2fc0de.mov

- Run yarn install and pod install in example to upgrade the react navigation dependencies.
- Add `fullScreenGestureEnabled` in react navigation screen options

- Open any pager view example and `swipe to go back` with `fullScreenGestureEnabled` should work on the first page.
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist


ios/Fabric/RNCPagerViewComponentView.mm
```diff
@@ -398,6 +406,29 @@ + (ComponentDescriptorProvider)componentDescriptorProvider
 }
 
 
+- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecognizeSimultaneouslyWithGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer {
+
+    // Recognize simultaneously only if the other gesture is RN Screen's pan gesture (one that is used to perform fullScreenGestureEnabled)
+    if (gestureRecognizer == self.panGestureRecognizer && [NSStringFromClass([otherGestureRecognizer class]) isEqual: @"RNSPanGestureRecognizer"]) {
```
intergalacticspacehighway, line 412: `RNSPanGestureRecognizer` is the name of the pan gesture used by [react-native-screens](https://github.com/software-mansion/react-native-screens/blob/4438732600e2d47105c4ffd9e62b33ef8723430e/ios/RNSScreenStack.mm#L82). I couldn't find a better approach than this, as [discussed here](https://github.com/callstack/react-native-pager-view/pull/665#discussion_r1038906896)

andreialecu:
Awesome, thanks for this!

One quick question though - does this only apply to the first page?

If the user drags from the very left edge while on pane 2 of the pager, will he be allowed to go back in React Navigation on iOS?

https://user-images.githubusercontent.com/697707/227491153-d4fc852c-3972-4287-a1c4-bb728e998a15.mp4

Here's what I mean, above. Right now we're doing this via a negative hitSlop as per this comment: https://github.com/PedroBern/react-native-collapsible-tab-view/issues/248#issuecomment-1081087355

However it doesn't seem to always work, and has the side effect of nothing being able to be clicked within that edge.

intergalacticspacehighway:
> from the very left edge while on pane 2 of the pager
Yes, it should work. It at least works in the example app. It needs native stack navigator though. 

https://user-images.githubusercontent.com/23293248/227498058-698c9151-ffe3-42f4-8580-535edc6dc24c.mov

btw I noticed the scroll view based rewrite on [next](https://github.com/callstack/react-native-pager-view/tree/next) branch. So would need a different approach to make it work in that version. I can look when I get some time 😅

intergalacticspacehighway:
@andreialecu are you using native stack? if yes, I can look into it if you can share a minimal repro. JS based Stack navigator doesn't work with the edge gesture and needs the hitSlop workaround like you mentioned.

andreialecu:
Ah, not on the native stack yet because we still rely on https://github.com/IjzerenHein/react-native-shared-element which doesn't work with the native stack. 

Once the shared element transitions in Reanimated mature, we should be able to move to the native stack finally. They're currently still incomplete and buggy.

Thanks for clarifying!

troZee:
@intergalacticspacehighway @andreialecu 
Thank you for sharing your thoughts here. I have tested this PR and it works 🎉

> btw I noticed the scroll view based rewrite on [next](https://github.com/callstack/react-native-pager-view/tree/next) branch.

will create another ticket for that to have visibility on it.

troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v6.2.0

ouabing:
I used to encountered an issue after applying this patch in the previous PR with react-native-screens enabled.

After I navigate to another screen and go back, `self.panGestureRecognizer` is changed and `shouldRecognizeSimultaneouslyWithGestureRecognizer` seems to not correctly recognize the latest self.panGestureRecognizer.

My observation is `didMoveToWindow` get called everytime I reenter the screen with PagerView if `react-native-screens` is enabled , and [this line](https://github.com/callstack/react-native-pager-view/pull/705/files#diff-1e1d39da2442ac68ad4df58e06e8eeee97e8d63a55bf47a64f72139d9755e1ceR51) always assigns a new `UIPanGestureRecognizer` without remove the previous one.

My personal patch based on this PR is like below:

```
if (self.panGestureRecognizer) {
    [self removeGestureRecognizer:self.panGestureRecognizer];
}
UIPanGestureRecognizer* panGestureRecognizer = [UIPanGestureRecognizer new];
self.panGestureRecognizer = panGestureRecognizer;
panGestureRecognizer.delegate = self;
[self addGestureRecognizer: panGestureRecognizer];
```

Because I'm not sure if it's a common issue, so I just post here if someone need this.

intergalacticspacehighway:
@ouabing i moved the gesture initialiser from `didMoveToWindow` to `initWithEventDispatcher` in this PR. So, it should only get called once during the View creation. Can you try the version 6.2.0 and let us know if the issue persists? Thanks!

ouabing:
@intergalacticspacehighway oh my apology for not checking the PR carefully. I'll try now, thanks for the work!

ouabing:
Hi I just tested this PR and now I think my problem should be a new issue. I used to use this PR to solve another issue with `react-native-tab-view`(which uses react-native-pager-view), when a tab view is inside a pushed screen(not RNScreen's native stack but with RNScreen enabled with react-navigation), the back gesture won't work.

I noticed in the new PR the `otherGestureRecognizer` check is limited to the one RNScreen's native stack use, so the PR is not working any more in my case.

I will try to reproduce my case and create another issue maybe, thank you anyway!

boboxiaodd:
@troZee  i use react-native-pager-view for video show like TikTok
upgrade to 6.2.0 ,  gestureEnabled can't work for iOS ,i has to downgrade to 6.1.4

intergalacticspacehighway:
@boboxiaodd what do you mean `gestureEnabled` can't work for iOS? Can you explain in more detail or with a video? 

There is no prop named `gestureEnabled` in the library or do you mean `scrollEnabled`?

boboxiaodd:
@intergalacticspacehighway 
`gestureEnabled` is  [reactnavigation screen prop](https://reactnavigation.org/docs/stack-navigator/#gestureenabled)

tiagotwistag:
> Hi I just tested this PR and now I think my problem should be a new issue. I used to use this PR to solve another issue with `react-native-tab-view`(which uses react-native-pager-view), when a tab view is inside a pushed screen(not RNScreen's native stack but with RNScreen enabled in react-navigation), the back gesture won't work.
> 
> I noticed in the new PR the `otherGestureRecognizer` check is limited to the one RNScreen's native stack use, so the PR is not working any more in my case.
> 
> I will try to reproduce my case and create another issue maybe, thank you anyway!

I believe I'm also facing the same issue, Is there any plan to support @react-navigation/stack?

troZee:
Since the fact, this feat does not work, I am reverting it here: https://github.com/callstack/react-native-pager-view/pull/771

PR #711, opened 2023-03-27
Fix safe area in examples

jaworek:
# Summary

Fixes incorrect safe area for iOS examples.
Aditionally, updates "@types/react" as it was on older version which resulted in type errors. iOS pod for react-native-pager-view was bumped to latest version.

## Test Plan

Check if modified examples are working correctly on Android and iOS

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/example/src/tabView/TabBarIconExample.tsx
+++ b/example/src/tabView/TabBarIconExample.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
-import { StyleSheet } from 'react-native';
+import { StyleSheet, View } from 'react-native';
+import { useSafeAreaInsets } from 'react-native-safe-area-context';
 import {
   TabView,
   TabBar,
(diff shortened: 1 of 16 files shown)


example/src/tabView/TabBarIconExample.tsx
```diff
@@ -13,57 +14,61 @@ import Contacts from './Shared/Contacts';
...
 
 type State = NavigationState<Route>;
 
-export default class TabBarIconExample extends React.Component<{}, State> {
-  static title = 'Top tab bar with icons';
-  static backgroundColor = '#e91e63';
-  static appbarElevation = 0;
+export default function TabBarIconExample() {
+  const [index, setIndex] = React.useState(0);
+  const routes = [
```
okwasniewski, line 24: Same thing, let's move this outside of the component. You can also copy this from react navigation repo: https://github.com/react-navigation/react-navigation/tree/main/example/src/Screens/TabView

jaworek:
@okwasniewski I applied your suggestions.
@krozniata I checked fabric examples and it seems that they are nearly identical, so I applied these fixes to them as well.

krozniata (approved):

okwasniewski (approved):

Issue #709, opened 2023-03-24
[iOS] back gesture in 7.0.0


troZee:
> from the very left edge while on pane 2 of the pager

Yes, it should work. It at least works in the example app. It needs native stack navigator though. 

https://user-images.githubusercontent.com/23293248/227498058-698c9151-ffe3-42f4-8580-535edc6dc24c.mov

btw I noticed the scroll view based rewrite on [next](https://github.com/callstack/react-native-pager-view/tree/next) branch. So would need a different approach to make it work in that version. I can look when I get some time 😅

_Originally posted by @intergalacticspacehighway in https://github.com/callstack/react-native-pager-view/issues/705#issuecomment-1482586811_

PR #712, opened 2023-03-28
fix(ios): v7 swipe to go back gestures with react navigation

intergalacticspacehighway:
# Summary
Resolves - https://github.com/callstack/react-native-pager-view/issues/709

- Adds swipe-to-go-back from the edge and `fullScreenGestureEnabled` support in scroll view implementation of pager view.
- This has a much cleaner approach thanks to the new scroll view based implementation, we don't need an extra pan gesture.

## Test Plan
Test example and fabric example with native stack and `fullScreenGestureEnabled`

Run `yarn install` and `pod install` in example and fabric example
- Test if swipe to go back gesture from edge works
- Add `fullScreenGestureEnabled` and test if full screen swipe from the first page pops the screen.
## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist


intergalacticspacehighway:
open to changing the name of `RNCPageScrollView` since it's not an actual React Native Component. Couldn't decide so kept this for now 😅. Maybe just `PagerViewScrollView`?

ios/RNCPagerScrollView.m
```diff
@@ -0,0 +1,27 @@
+
+#import "RNCPagerScrollView.h"
+
+@implementation RNCPagerScrollView
+- (instancetype)initWithFrame:(CGRect)frame {
+    self = [super initWithFrame:frame];
+    if (self) {
```
okwasniewski, line 7: Can you change this to 

```obj-c
if (self = [super initWithFrame:frame]) {
```

Just to have consistency across files.

ios/RNCPagerScrollView.m
```diff
@@ -0,0 +1,27 @@
...
+        self.panGestureRecognizer.delegate = self;
+    }
+    return self;
+}
+
+- (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecognizeSimultaneouslyWithGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer {
```
okwasniewski, line 13: Also how about moving this delegate to an extension? It will make the code a bit cleaner.

okwasniewski (commented):
Thanks for this! 🎉 I'm leaving two minor comments and I think we will be ready to merge.

ios/RNCPagerScrollView.m
intergalacticspacehighway: I am a bit obj c noob. Do you mean to add a category and make a separate `RNCPagerScrollView+UIGestureRecognizerDelegate.m` file?

krozniata (approved):

ios/RNCPagerScrollView.m
okwasniewski: Let's merge it as is since it's minimal change.

okwasniewski (approved):

PR #715, opened 2023-04-04
Fix issues reported by QA

jaworek:
# Summary

- Fix duplicated key in logs panel
- Use `TouchableOpacity` from RN instead of react-native-gesture-handler, on Android it did not work. It was caused by missing setup from these docs: https://docs.swmansion.com/react-native-gesture-handler/docs/installation#usage-with-modals-on-android . Based on the info from @troZee we don't need it, so I changed it to one provided by RN.
- Add missing testId to simplify tests written in Maestro
- Prevent removing last page

## Test Plan

1. Validate that events generated in logs panel do not trigger duplicated key warning.
2. Verify that control tab works correctly on Android.
3. Verify that removing last page does not remove page if only 1 is present

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/example/src/component/NavigationPanel/NavigationPanel.tsx
+++ b/example/src/component/NavigationPanel/NavigationPanel.tsx
@@ -48,7 +48,6 @@ export function NavigationPanel(props: NavigationPanelProps) {
           ]}
           activeOpacity={0.8}
           onPress={() => {
-            console.log('edek');
             setVisible((prevVisible) =>
               prevVisible === VisibleTab.Logs
                 ? VisibleTab.None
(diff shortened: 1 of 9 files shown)


example/src/component/NavigationPanel/NavigationPanel.tsx
```diff
@@ -48,13 +47,14 @@ export function NavigationPanel(props: NavigationPanelProps) {
             visible === VisibleTab.Logs && styles.toggleVisibilityButtonActive,
           ]}
           activeOpacity={0.8}
-          onPress={() =>
+          onPress={() => {
+            console.log('edek');
```
krozniata, line 51: edek 💯  can be removed

Issue #722, opened 2023-05-03
6.2.0 breaks vertical scrolling of ScrollView parent


jwh-hutchison:
## Environment

"react-native": "0.71.6",
"react-native-pager-view": "6.2.0",
"react-native-tab-view": "^3.5.1",

## Description

I was trying to render `TabView` of react-native-tab-view inside of a `ScrollView`, but after upgrading to v6.2.0 for pager-view, the vertical scrolling of the scroll view would lock and only adjust if I didn't select the element that was being rendered as a child under the `render()` function of the PagerViewAdapter; so if I tried scrolling when selecting a `TabBar` it would scroll correctly - if I switched out the `render(<SceneView/>)` for a generic `View` I could scroll fine.

Inside react-native-tab-view/src/TabView:

```
return (
  <React.Fragment>
    {tabBarPosition === 'top' &&
      renderTabBar({
        ...sceneRendererProps,
        navigationState,
      })}
    {render(
      navigationState.routes.map((route, i) => {
        return (
          <SceneView
            {...sceneRendererProps}
            addEnterListener={addEnterListener}
            key={route.key}
            index={i}
            lazy={typeof lazy === 'function' ? lazy({ route }) : lazy}
            lazyPreloadDistance={lazyPreloadDistance}
            navigationState={navigationState}
            style={sceneContainerStyle}
          >
            {({ loading }) =>
              loading
                ? renderLazyPlaceholder({ route })
                : renderScene({
                    ...sceneRendererProps,
                    route,
                  })
            }
          </SceneView>
        );
      })
    )}
    {tabBarPosition === 'bottom' &&
      renderTabBar({
        ...sceneRendererProps,
        navigationState,
      })}
  </React.Fragment>
);
```

This PagerView adapter inherits `PagerViewProps` props from react-native-pager-view and after downgrading the library (to "react-native-pager-view": "6.1.4"), the entire page can scroll fine again, I was originally going to pin the bug on react-native-tab-view but it seems that this issue has happened with other components that depend on the library (see: https://github.com/callstack/react-native-pager-view/issues/713)

## Reproducible Demo

When you use the following code and try to scroll vertically on any elements rendered inside of `_renderTab` nothing will happen. If you downgrade react-native-pager-view to 6.1.4 it will work.

```
import React, {Component} from 'react';
import {TabBar, TabView} from 'react-native-tab-view';

_renderTab = (tab) => {
  return (
    <View>{...some stuff to render}</View>
  )
}

[...]

PR #734, opened 2023-06-05
fix(ios): parent scroll view not able to scroll

intergalacticspacehighway:
# Summary

Fixes - https://github.com/callstack/react-native-pager-view/issues/722

The default iOS's ScrollView allows nested ScrollView. i.e. a child scroll view doesn't prevent scrolling of the parent scroll view. It's done internally but I couldn't find any documentation on how it's being done, the delegates used by ScrollView doesn't seem to be public. Since we're overriding shouldRecognizeSimultaneouslyWithGestureRecognizer and returning NO, the parent scroll view is not able to scroll anymore. I set it to NO because that is the default behaviour in most cases but looks like to support Parent scroll views we need to return YES when the other recognizer is a ScrollView. This surely needs some more testing though.

## Test Plan
In the below code snippet, the parent scroll view should be able to scroll.

```jsx
import React, { Component } from 'react';
import { ScrollView, View, Text } from 'react-native';
import { TabBar, TabView } from 'react-native-tab-view';

export class Navigation extends Component {
  state = {
    index: 0,
    routes: [
      { key: 'day', title: 'Day' },
      { key: 'week', title: 'Week' },
      { key: 'month', title: 'Month' },
    ],
  };

  _renderTab = (tab) => {
    return <View style={{ backgroundColor: 'pink' }}></View>;
  };

  _renderTabBar = (props) => {
    return (
      <TabBar
        {...props}
        renderLabel={({ route }) => (
          <View>
            <Text style={{}}>{route.title}</Text>
          </View>
        )}
      />
    );
  };

  _renderScene = ({ route, jumpTo }) => {
    switch (route.key) {
      case 'day':
        return this._renderTab('day');
      case 'week':
        return this._renderTab('week');
      case 'month':
        return this._renderTab('month');
    }
  };

  render() {
    return (
      <ScrollView
        contentContainerStyle={{ flexGrow: 1, marginTop: 44 }}
        nestedScrollEnabled={true}
        scrollEnabled={true}
      >
        <View style={{ flex: 1 }}>
          <TabView
            navigationState={this.state}
            onIndexChange={(index) => {
              this.setState({ index });
            }}
            renderScene={this._renderScene}
            renderTabBar={this._renderTabBar}
          />
        </View>
      </ScrollView>
    );
  }
}

```

- Just copy/paste the above snippet in App.tsx

- Before the native changes in the PR, the parent scroll view won't work. Adding the native changes from the PR should fix it.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/Fabric/RNCPagerViewComponentView.mm
+++ b/ios/Fabric/RNCPagerViewComponentView.mm
@@ -408,7 +408,7 @@ using namespace facebook::react;
 
 - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecognizeSimultaneouslyWithGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer {
 
-    // Recognize simultaneously only if the other gesture is RN Screen's pan gesture (one that is used to perform fullScreenGestureEnabled)
+    // Recognize simultaneously if the other gesture is RN Screen's pan gesture (one that is used to perform fullScreenGestureEnabled)
     if (gestureRecognizer == self.panGestureRecognizer && [NSStringFromClass([otherGestureRecognizer class]) isEqual: @"RNSPanGestureRecognizer"]) {
         UIPanGestureRecognizer* panGestureRecognizer = (UIPanGestureRecognizer*) gestureRecognizer;
         CGPoint velocity = [panGestureRecognizer velocityInView:self];
@@ -424,6 +424,12 @@ using namespace facebook::react;
         
         return YES;
     }
+    
+    // Allow nested scroll views to scroll simultaneously with the pager
+    if ([otherGestureRecognizer.view isKindOfClass: UIScrollView.class]) {
+        return YES;
+    }
+
     const auto &viewProps = *std::static_pointer_cast<const RNCViewPagerProps>(_props);
     scrollView.panGestureRecognizer.enabled = viewProps.scrollEnabled;
     return NO;
--- a/ios/ReactNativePageView.m
+++ b/ios/ReactNativePageView.m
@@ -468,7 +468,7 @@
 
 - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecognizeSimultaneouslyWithGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer {
 
-    // Recognize simultaneously only if the other gesture is RN Screen's pan gesture (one that is used to perform fullScreenGestureEnabled)
+    // Recognize simultaneously if the other gesture is RN Screen's pan gesture (one that is used to perform fullScreenGestureEnabled)
     if (gestureRecognizer == self.panGestureRecognizer && [NSStringFromClass([otherGestureRecognizer class]) isEqual: @"RNSPanGestureRecognizer"]) {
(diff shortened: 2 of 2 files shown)


voloshinskii:
Hi, @intergalacticspacehighway. Thank you for your PR! 

Now, in our project, we still have several issues with that fix
- In modal presentationStack PagerView breaks modal swipe-out
- Is there any reason to block other recognizers when `self.scrollEnabled` equals false?

ios/Fabric/RNCPagerViewComponentView.mm
```diff
@@ -408,7 +408,7 @@ + (ComponentDescriptorProvider)componentDescriptorProvider
 
 - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecognizeSimultaneouslyWithGestureRecognizer:(UIGestureRecognizer *)otherGestureRecognizer {
```
voloshinskii, line 410:
```suggestion
    if (!self.scrollEnabled) {
        return YES;
    }
```

voloshinskii: I guess, we should always recognize simultaneously in that case

intergalacticspacehighway: simultaneous by default [returns false](https://developer.apple.com/documentation/uikit/uigesturerecognizerdelegate/1624208-gesturerecognizer#return_value). So I wanted to preserve the original behaviour and allow recognize only when it's RN screen's triggered pop gesture. But i think your suggestion might also make sense e.g. current gesture is blocked so allow the other 🤔. Does this change fixes your issues?

troZee:
Hey 👋 
It seems you would like to achieve the same effect as in this library https://github.com/PedroBern/react-native-collapsible-tab-view. I would prefer to contribute to that library since they support the parallax effect. Right now, I am closing this issue, since it is not supported in the pager view.

interstates21:
Would be great to have it fixed though

FlashGameDev:
Why is this closed without being merged in? This is still an issue in 6.6.1

Dat-Mobile:
I'm so sick of it, even `react-native-pager-view` version `6.6.1` still hasn't been fixed (Jan 8, 2025). The reason I'm not using 7.0.0 is because it breaks the build for New architecture
Here is the patch for the fix of `6.6.1`, please feel free to use

```
index 652a5c1..44a90eb 100644
--- a/node_modules/react-native-pager-view/ios/Fabric/RNCPagerViewComponentView.mm
+++ b/node_modules/react-native-pager-view/ios/Fabric/RNCPagerViewComponentView.mm
@@ -439,6 +439,13 @@ - (BOOL)gestureRecognizer:(UIGestureRecognizer *)gestureRecognizer shouldRecogni
         
         return YES;
     }
+
+
+    // Allow nested scroll views to scroll simultaneously with the pager
+    if ([otherGestureRecognizer.view isKindOfClass: UIScrollView.class]) {
+        return YES;
+    }
+    
     const auto &viewProps = *std::static_pointer_cast<const RNCViewPagerProps>(_props);
     scrollView.panGestureRecognizer.enabled = viewProps.scrollEnabled;
     return NO;
```

PR #736, opened 2023-06-19
feat(android): 🌟 add support for React Native 0.73

MasonLe2497:
# Summary
Starting from React Native v0.73 , all libraries will need to be updated with these two one-liners due to Android Gradle Plugin upgrade
[discussions-671](https://github.com/react-native-community/discussions-and-proposals/issues/671)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -34,7 +34,7 @@ def getExtOrIntegerDefault(name) {
   return rootProject.ext.has(name) ? rootProject.ext.get(name) : (project.properties['PagerView_' + name]).toInteger()
 }
 def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0].toInteger()
-def shouldUseNameSpace = agpVersion >= 7
+def shouldUseNameSpace = agpVersion >= 8
 def PACKAGE_PROP = "package=\"com.reactnativepagerview\""
 def manifestOutFile = file("${projectDir}/src/main/AndroidManifest.xml")
 def manifestContent = manifestOutFile.getText()
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -36,7 +36,11 @@ def getExtOrIntegerDefault(name) {
 
 android {
   compileSdkVersion getExtOrIntegerDefault('compileSdkVersion')
-  namespace = "com.reactnativepagerview"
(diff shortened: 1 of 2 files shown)


troZee:
Hello 👋 
Thank you for providing a pr. Unfortunately, it will break apps < 71( https://github.com/react-native-community/discussions-and-proposals/issues/671#issuecomment-1597386954)

```
 it's better to don't remove the package directive from the Manifest and add the namespace directive.
 ``` 
 
 Please update your PR regarding the above comment.  Thank you 🙏

**Review by troZee — CHANGES_REQUESTED** [review]
https://github.com/callstack/react-native-pager-view/pull/736#issuecomment-1597396610

**Comment by MasonLe2497** [comment]
@troZee i updated my PR

File: android/build.gradle [hunk 1]
```diff
@@ -33,10 +33,32 @@ def getExtOrDefault(name) {
 def getExtOrIntegerDefault(name) {
   return rootProject.ext.has(name) ? rootProject.ext.get(name) : (project.properties['PagerView_' + name]).toInteger()
 }
+def agpVersion = com.android.Version.ANDROID_GRADLE_PLUGIN_VERSION.tokenize('.')[0].toInteger()
```
troZee, line 36: Could you explain the responsibilities of this code?

MasonLe2497: RN 0.71 depended on classpath("com.android.tools.build:gradle:7.4.1"). So i get major version of build:gradle  to add namespace and remove package in AndroidManifest.

android/build.gradle
```diff
@@ -33,10 +33,32 @@ def getExtOrDefault(name) {
...
+def shouldUseNameSpace = agpVersion >= 7
```
atlj, line 37: Namespace support was added with Android Gradle Plugin 7.3, [source](https://developer.android.com/build/releases/past-releases/agp-8-0-0-release-notes#namespace-dsl)

This comparison would break builds with AGP version `7.2`, and so on.
So I would suggest to ignore version 7 and only add the namespace property on version 8.
```suggestion
def shouldUseNameSpace = agpVersion >= 8
```

troZee (approved):

PR #771, opened 2023-09-13
Revert "feat(ios): back gesture with full screen swipe (#705)"

td-tomasz-joniec:
This reverts commit 769a3a21500fbe40c1909c9bf50f724c76a5b435.

# Summary

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/ios/Fabric/RNCPagerViewComponentView.mm
+++ b/ios/Fabric/RNCPagerViewComponentView.mm
@@ -15,10 +15,7 @@
 
 using namespace facebook::react;
 
-@interface RNCPagerViewComponentView () <RCTRNCViewPagerViewProtocol, UIPageViewControllerDataSource, UIPageViewControllerDelegate, UIScrollViewDelegate, UIGestureRecognizerDelegate>
-
-@property(nonatomic, assign) UIPanGestureRecognizer* panGestureRecognizer;
-
+@interface RNCPagerViewComponentView () <RCTRNCViewPagerViewProtocol, UIPageViewControllerDataSource, UIPageViewControllerDelegate, UIScrollViewDelegate>
 @end
 
 @implementation RNCPagerViewComponentView {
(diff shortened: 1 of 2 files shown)


zRelux:
+ 1 this made it stop working for me as well with `react-native-collapsible-tab-view`

troZee:
@intergalacticspacehighway Can we revert your changes? Maybe we should consider adding this fix https://github.com/callstack/react-native-pager-view/pull/705#issuecomment-1483855040 


@td-tomasz-joniec what functionality will be broken, when this patch is removed? Will it introduce some regression?

intergalacticspacehighway:
@zRelux what has stopped working? Can you please explain the issue?

@troZee reverting can break the back gesture when `fullScreenSwipeEnabled` is enabled for pop react-navigation. Not a massive breaking change, but I am still unclear on the reason for reverting.

The additional issue asked [here ](https://github.com/callstack/react-native-pager-view/pull/705#issuecomment-1483869102) is a different feature request to support non-native stack (allow swiping back from the edge when currentPage > 0 and using non-native stack). I can look into that but that won't be solved even if we revert this PR.

td-tomasz-joniec:
> @intergalacticspacehighway Can we revert your changes? Maybe we should consider adding this fix [#705 (comment)](https://github.com/callstack/react-native-pager-view/pull/705#issuecomment-1483855040)
> 
> @td-tomasz-joniec what functionality will be broken, when this patch is removed? Will it introduce some regression?

I am only interested in reverting this change as this code breaks correct touch handling when pager view is inside scrollview

intergalacticspacehighway:
@td-tomasz-joniec are you facing this https://github.com/callstack/react-native-pager-view/issues/722 issue?

troZee (changes requested):
Can we add this code based on prop called `rnsPanGestureRecognizer`?

troZee:
> Any update ? still not working for me even on 6.1.4, I'm using @react-navigation non native Stack and react-native-collapsible-tab-view @intergalacticspacehighway

Could you provide a repro? It will definitely speed it up. Right now, I cannot reproduce it

IN-applaunch:
I'm facing an issue with gesture detection in my React Native project. I'm implementing a custom tab view and also using a pager component. The problem arises because, when using these together, the gesture detection doesn't work as expected. It seems like one gesture handler is overriding or interfering with the others, leading to some gestures not being detected at all. This results in interaction issues, where users can't swipe or interact with the views properly.

The specific problem is that the pager's gesture detection takes priority, which then disables or limits the gestures in my custom tab view. I'm looking for a solution that allows both components to handle gestures independently without conflict, ensuring smooth and expected behavior in my app's UI.

PR #775, opened 2023-10-11
chore(next): upgrade RN version in both examples

krozniata:
# Summary

This PR upgrades RN version to 72.5 in both **Paper** and **Fabric** example.

It also fixes some smaller issues with building and android version getting stuck after `setPage` call in Fabric version

## Test Plan

Both Paper and Fabric examples should work correctly on both platforms

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist


example/android/gradle.properties
```diff
@@ -25,7 +25,7 @@ android.useAndroidX=true
 android.enableJetifier=true
 
 # Version of flipper SDK to use with React Native
-FLIPPER_VERSION=0.174.0
+FLIPPER_VERSION=0.182.0
```
okwasniewski, line 28: Is this needed? As we are not using flipper and it's getting deprecated

krozniata: I just followed RN upgrade helper 😄 I guess it will be soon removed in 0.73

okwasniewski (approved):

PR #778, opened 2023-10-13
feat(next): remove fabric example

krozniata:
# Summary

This PR removes fabric example from `next` branch


README.md
```diff
@@ -41,29 +34,6 @@ In version **6.x** support for `transitionStyle` property has been dropped. More
 
 `yarn add react-native-pager-view`
 
-## New architecture setup (Fabric)
-
-This library supports new architecture! We have two example folders one for each architecture. If you are using this library in your own project there some extra steps needed.
```
okwasniewski, line 46: I would leave this info and update the instructions on how to run normal example with fabric enabled

package.json
```diff
@@ -30,8 +30,6 @@
     "releaseNext": "release-it --npm.tag=next --preRelease=rc",
     "example": "yarn --cwd example",
     "bootstrap": "yarn example && yarn && yarn example pods",
-    "fabricexample": "yarn --cwd fabricexample",
-    "bootstrap-fabric": "yarn fabricexample && yarn && yarn fabricexample pods",
```
okwasniewski, line 34: How about leaving this `bootstrap-fabric` script and making it actually bootstrap fabric in the `example` folder?

README.md
krozniata: I removed it so it matches current README on `main` branch, but I think it will be good idea to create a PR adding fabric instructions to `main` branch and just sync them. Same applies for comment about `yarn bootstrap-fabric` because `main` is also missing that script

okwasniewski: Okay, let's do it in a separate PR

okwasniewski (approved):

PR #793, opened 2023-11-17
fix(next/iOS): fix blinking during initial render when initialPage !=0 in non-fabric version

ku8ar:
# Summary

Fabric has a trick with -1. Unfortunately here we can't use it. The only place that allows a valid goTo from the initial page is layoutSubviews.

## Test Plan

1. render pager with initialPage bigger than zero and many pages inside
2. pager is blinking (for a brief moment)

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/RNCPagerView.m
+++ b/ios/RNCPagerView.m
@@ -17,6 +17,7 @@
 
 @property(nonatomic, strong) RNCPagerScrollView *scrollView;
 @property(nonatomic, strong) UIView *containerView;
+@property(nonatomic, assign) BOOL scrolledToInitialPage;
 
 - (void)goTo:(NSInteger)index animated:(BOOL)animated;
 - (void)shouldScroll:(BOOL)scrollEnabled;
@@ -46,6 +47,15 @@
         [self.scrollView.panGestureRecognizer requireGestureRecognizerToFail:self.reactViewController.navigationController.interactivePopGestureRecognizer];
     }
 }
+
+- (void)layoutSubviews {
+    [super layoutSubviews];
+    if (!_scrolledToInitialPage) {
+        _scrolledToInitialPage = YES;
+        [self goTo:self.initialPage animated:false];
+    }
+}
+
 
 - (void)didUpdateReactSubviews {
     [self updateContentSizeIfNeeded];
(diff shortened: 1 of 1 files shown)


krozniata (commented):
Hey @ku8ar! 
Could you provide video showing blinking issue? I couldn't reproduce it
Also that change introduced a bug causing initial page not to be set correctly once every few render (`initialPage` set to  e.g.`7` with `100` pages in `BasicPagerViewExample`)

ku8ar:
Thats weird what you say. Maybe you accidentally ran the fabric version? Here is video that demonstrates problem (full-screen video is rendered inside pager-view):

https://github.com/callstack/react-native-pager-view/assets/16975059/e60f1e25-024d-4dbd-a63c-1288a18303a1

After this patch, initial page is always rendered correctly when pager-view is loaded:

https://github.com/callstack/react-native-pager-view/assets/16975059/2c4d2e77-7e17-44f5-99b6-5b4962123a60


Screen/pager-view rendering code in our app has not been changed since version 6.x.

ios/RNCPagerView.m
```diff
@@ -47,6 +48,15 @@ - (void)didMoveToWindow {
...
 
+- (void)layoutSubviews {
+    [super layoutSubviews];
+    if (!_scrolledToInitialPage) {
+        _scrolledToInitialPage = YES;
+        [self goTo:self.initialPage animated:false];
+    }
+}
+
+
```
krozniata, line 59: Looks like bug causing `initialPage` sometimes not to be set correctly can be easily fixed by adding check for `contentSize` like: 
```objc
-(BOOL)isScrollViewContentSizePositive {
    return [self isHorizontal] ? _scrollView.contentSize.width > 0 : _scrollView.contentSize.height > 0;
}
```
and adding it to

```objc
-(void)layoutSubviews {
     [super layoutSubviews];
     if (self.isScrollViewContentSizePositive && !_scrolledToInitialPage) {
         _scrolledToInitialPage = YES;
         [self goTo:self.initialPage animated:false];
     }
 }
 ```
 
 Could you please add it?

PR #802, opened 2023-12-21
feat: added sensitivity controller to android

adamsythe:
# Summary

This pull request allows users to toggle the sensitivity on android.  It is particularly useful for pagerview's that contain scrollable webviews.  

## Test Plan

Make sure the prop filters through the JS to reduceDragSensitivity kotlin function and is set correctly.  Make sure the app visibly scrolls with less sensitivity as the number increasers.

Just be able to run the app in android studio

Go to BasicPagerViewExample and add the prop
ScrollSensitivity={8}
in Android studio put a stop on 
android/src/main/java/com/reactnativepagerview/PagerViewViewManagerImpl.kt on line 25.

Run the basic example and when the stop fires check that the prop is correctly set at 8.

You can also toggle the number and notice the reduced sensitivity on higher numbers.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅    |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativepagerview/PagerViewViewManagerImpl.kt
+++ b/android/src/main/java/com/reactnativepagerview/PagerViewViewManagerImpl.kt
@@ -3,6 +3,7 @@ package com.reactnativepagerview
 import android.view.View
 import androidx.viewpager2.widget.ViewPager2
 import com.facebook.react.uimanager.PixelUtil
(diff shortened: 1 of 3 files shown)


MrRefactor:
Hey, as this PR is adding sensitivity controller to android only, Im closing it as we want to have feature parity on both iOS and android. Thanks for the effort!

PR #804, opened 2024-01-18
fix(ios): fix freezing when navigating to same index

### LINKED ISSUE #803: Tapping twice on tab freezes tab bar - iOS - 7.0.0-rc.1
## Environment

iOS 17.2
react-native-pager-view: v7.0.0-rc.1
react-native-tab-view: 3.3.0

## Description

Tapping twice on a tab freezes the tab bar. This is due to the `RNCPagerView` implementation setting `self.animating = true;` in the `goTo` method, which is never reset back to `false` when you tap on the same tab twice.

## Reproducible Demo

Tap twice on a tab and see the tab bar frozen.

## Potential fix

I believe `self.animating = true;` could be conditionally set only when the index and current page differ i.e. `[self getCurrentPage] != index` in the `goTo` method.

krozniata:
# Summary

This PR fixes issue causing `PagerView` to freeze after calling `goTo` to same index (Paper arch only)

Fixes #803 

## Test Plan

1. Open any example with tab
2. Tap on currently selected tab
3. Pager should be scrollable

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

## Checklist

### DIFF
--- a/ios/RNCPagerView.m
+++ b/ios/RNCPagerView.m
@@ -126,6 +126,10 @@
 }
 
 - (void)goTo:(NSInteger)index animated:(BOOL)animated {
+    if ([self getCurrentPage] == index) {
+        return;
+    }
+    
     CGPoint targetOffset = [self isHorizontal] ? CGPointMake(_scrollView.frame.size.width * index, 0) : CGPointMake(0, _scrollView.frame.size.height * index);
     
     if (animated) {


okwasniewski:
This looks similar to this one which had to be reverted: https://github.com/callstack/react-native-pager-view/pull/768/files

krozniata:
@okwasniewski It was for Fabric implementation

Here when calling `goTo` to same index, `scrollViewDidEndScrollingAnimation` is never called which causes `self.animating` to stay as true and it prevents `goTo` from being called (`L31` in `RNCPagerViewManager`)

Issue #823, opened 2024-05-03
[TS] PagerView typescript is broken


AlirezaHadjar:
## Description

I just noticed that in the latest version, the typescript of this package is not working as expected.

## Reproducible Demo

1. Create a ref:
```tsx
const ref = useRef<PagerView>(null);
```
Now you will get this ts error: `'PagerView' refers to a value, but is being used as a type here. Did you mean 'typeof PagerView'`

![Screenshot 2024-05-03 at 7 06 04 PM](https://github.com/callstack/react-native-pager-view/assets/57192409/7e541bf4-9f15-4312-9355-b606807c1929)

renchap:
I can reproduce this with 6.3.1, while it works with 6.3.0. 

https://github.com/callstack/react-native-pager-view/pull/817 has been merged and is in the `v6.3.1` tag which I think it a mistake, because it means that 6.3.1 is in equivalent to 7.0.0-rc2?

Pining @troZee here because I do not think this was your intent at all!

asusoft:
any fix??

sergeymild:
The same error

Jonnboy91:
I came across this same issue as well. Noticed that on iOS it still seemed to work even with the typescript complaints, but Android was not working anymore on the screens that uses this.

As mentioned previously locking the version to 6.3.0 works.

DrRefactor:
@troZee I'm not @MrRefactor but funny enough, "almost am" (look at the nicknames). Please see my PR:
https://github.com/callstack/react-native-pager-view/pull/827

renchap:
This is fixing `main`, but you still published a patch release (6.3.1) with the content of `main`, which makes it equivalent to a major 7.0.0 release.

gottfriedaniel:
When you'll release this fix?

PR #827, opened 2024-05-21
(#823) Export PagerView type

DrRefactor:
# Summary
Fixes https://github.com/callstack/react-native-pager-view/issues/823

React.forwardRef typing does not allow simple usage of component as a ref type. This PR brings backward compatibility of Ref<PagerView>.

## Test Plan

Test PagerView component, ref, props typing.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/src/PagerView.tsx
+++ b/src/PagerView.tsx
@@ -233,3 +233,9 @@ export const PagerView = React.forwardRef<PagerViewInternal, PagerViewProps>(
     return <PagerViewInternal {...rest} useLegacy={!useNext} ref={ref} />;
   }
 );
+
+// React.forwardRef does not type returned component properly, thus breaking Ref<MyComponent> typing.
+// One way to overcome this is using separate typing for component "interface",
+// but that breaks backward compatibility in this case.
+// Approach of merging type is hacky, but produces a good typing for both ref attributes and component itself.
+export type PagerView = PagerViewInternal & typeof PagerView;


src/PagerView.tsx
```diff
@@ -233,3 +233,9 @@ export const PagerView = React.forwardRef<PagerViewInternal, PagerViewProps>(
     return <PagerViewInternal {...rest} useLegacy={!useNext} ref={ref} />;
   }
 );
+
+// React.forwardRef does not type returned component properly, thus breaking Ref<MyComponent> typing.
+// One way to overcome this is using separate typing for component "interface",
+// but that breaks backward compatibility in this case.
+// Approach of merging type is hacky, but produces a good typing for both ref attributes and component itself.
+export type PagerView = PagerViewInternal & typeof PagerView;
```
DrRefactor, line 241: I'm addressing 2 problems here:
1. PagerView is a value, not a type, since we're using forwardRef. Therefore, it cannot be used as type at all, e.g. in `useRef<PagerView>`, current error is that PagerView is a value that's attempted to be used as a type. This could be mitigated by simply: `export type PagerView = typeof PagerView`. We can create types with the same names as values, TS chooses type/value basing on usage context (similar as built-in class-type mechanism in TS)
2. forwardRef does not type component "interface" at all. `typeof PagerView` does not show any ref based methods.

To overcome this, there are two approaches:
1. Export separate typing for component interface, e.g. `PagerViewInterface/Ref` and make all users switch to using it in useRef
2. The one I'm using above, that has a drawback of mixing two types, but preserves backward compatibility.

I suggest the second one, because lib user facing interface shouldn't be affected whether class or function component is used internally.

MrRefactor: @troZee

MrRefactor (approved):

troZee (approved):

PR #873, opened 2024-09-05
feat(js): add view pager hook

troZee:
# Summary

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅❌     |
| Android |    ✅❌     |

## Checklist

### DIFF
--- a/example/src/PagerHookExample.tsx
+++ b/example/src/PagerHookExample.tsx
@@ -49,6 +49,7 @@ export function PagerHookExample() {
           [rest.pages]
         )}
       </AnimatedPagerView>
+      {/*@ts-ignore*/}
       <NavigationPanel {...rest} />
     </SafeAreaView>
   );
--- a/example/src/PagerHookExample.tsx
+++ b/example/src/PagerHookExample.tsx
@@ -1,0 +1,70 @@
+import React, { useMemo } from 'react';
+import { StyleSheet, View, SafeAreaView, Text } from 'react-native';
+
+import { LikeCount } from './component/LikeCount';
+import { NavigationPanel } from './component/NavigationPanel';
+import { useViewPager } from 'react-native-pager-view';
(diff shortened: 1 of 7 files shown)


bataevvlad:
https://github.com/user-attachments/assets/6b49824e-ac5d-44ec-9fab-41b11ec0d79d

example/src/PagerHookExample.tsx
```diff
@@ -0,0 +1,71 @@
+import React, { useMemo } from 'react';
+import { StyleSheet, View, SafeAreaView, Text } from 'react-native';
+
+import { LikeCount } from './component/LikeCount';
+import { NavigationPanel } from './component/NavigationPanel';
+import { useViewPager } from 'react-native-pager-view';
+
+export function PagerHookExample() {
+  const { AnimatedPagerView, ref, ...rest } = useViewPager(10);
```
okwasniewski, line 9: I think in order to make this scalable and not introduce breaking changes in the future, a better design will be to use an object:

```ts
const { AnimatedPagerView, ref, ...rest } = useViewPager({pageCount: 10});
```

This way we can easily add more parameters later

src/useViewPager.ts
```diff
@@ -0,0 +1,140 @@
...
+              position: onPageScrollPosition,
+            },
+          },
+        ],
+        {
+          useNativeDriver: true,
+        }
+      ),
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+    []
```
okwasniewski, line 93: Why?

src/useViewPager.ts
```diff
@@ -0,0 +1,140 @@
+import type * as ReactNative from 'react-native';
```
troZee, line 1: Please rename it to `usePagerView`

PR #885, opened 2024-10-08
fix(ios): Fix vertical list scroll directionin rtl

DrRefactor:
# Summary

When in RTL, vertical pager view is scrolling backwards. This is unwanted behavior (only horizontal axis should be inverted in RTL).

## Test Plan

Test LTR & RTL in legacy pager view.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |

## Checklist

### DIFF
--- a/ios/LEGACY/Fabric/LEGACY_RNCPagerViewComponentView.mm
+++ b/ios/LEGACY/Fabric/LEGACY_RNCPagerViewComponentView.mm
@@ -309,10 +309,6 @@ using namespace facebook::react;
     return [_layoutDirection isEqualToString: @"ltr"];
 }
 
-- (BOOL)isHorizontalRtlLayout {
-    return self.isHorizontal && ![self isLtrLayout];
-}
-
 - (void)scrollViewDidScroll:(UIScrollView *)scrollView {
     CGPoint point = scrollView.contentOffset;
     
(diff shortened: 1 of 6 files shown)


troZee (changes requested):

troZee:
@DrRefactor could you provide any documentation for the RTL, when the component is scrollable vertically?

DrRefactor:
@troZee For vertically scrolled lists, there's basically no change in layout in RTL vs LTR. The only one difference in this case is when the vertically scrolled list is a grid, then the first element should be aligned to the right edge.
See: https://developer.apple.com/design/human-interface-guidelines/right-to-left

ios/LEGACY/Fabric/LEGACY_RNCPagerViewComponentView.mm
```diff
@@ -309,6 +309,10 @@ - (BOOL)isLtrLayout {
     return [_layoutDirection isEqualToString: @"ltr"];
 }
 
+- (BOOL)isHorizontalRtlLayout {
```
troZee, line 312: Could you extract this function to a separate file? It is used in two places.

troZee (changes requested):

ios/LEGACY/Fabric/LEGACY_RNCPagerViewComponentView.mm
DrRefactor: Good point. I've done this in separate commit, as it introduces some boilerplate obj-c code. I'm more of an expert in swift than obj-c, so please verify.

MrRefactor (approved):

PR #915, opened 2024-11-19
fix(iOS): fix component layout issue

krozniata:
# Summary

This PR fixes issue with views not being correctly recycled on new arch, causing components styles miscalculations

## Test Plan

1. Open example app
2. Navigate through examples a lot – e.g `BasicExample` -> `PaginationDotsExample` -> `HeadphonesExample` -> again `PaginationDotsExample` etc.
3. No layout issues should appear, rest should work as before

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/example/src/HeadphonesCarouselExample.tsx
+++ b/example/src/HeadphonesCarouselExample.tsx
@@ -261,9 +261,10 @@ export default function HeadphonesCarouselExample() {
           }
         )}
       >
-        {data.map((item, index) => (
+        {data.map(({ key, ...item }, index) => (
           <View collapsable={false} key={index}>
             <Item
+              key={key}
               {...item}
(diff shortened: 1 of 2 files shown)


cipolleschi (approved):
The code makes sense to me. I think it is also what we have to do in RNGesture Handler to fix some issues: https://github.com/software-mansion/react-native-gesture-handler/pull/2926

Out of curiosity, why we have `ios/LEGACY/Fabric` path? Do we have a non legacy one? Which one is used in prod?

troZee (approved):

MrRefactor (approved):
lgtm

MrRefactor:
@cipolleschi released in https://github.com/callstack/react-native-pager-view/releases/tag/v6.5.1

PR #918, opened 2024-11-21
refactor(lib): Remove experimental Next implementation

MrRefactor:
# Summary
Remove experimental `scrollview` based `react-native-pager-view` implementation.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ✅     |

## Checklist

- [x] I have tested this on a device and a simulator
- [x] I added the documentation in `README.md`
- [x] I updated the typed files (TS and Flow)

PR #919, opened 2024-11-21
fix(#722): revert not working back navigation gesture in favour of PagerView inside ScrollView

### LINKED ISSUE #722: 6.2.0 breaks vertical scrolling of ScrollView parent

MateWW:
# Summary

#### Issue
Currently scroll over legacy pager view is not working while nested inside scroll view.
Nesting might be necessary while we have big header which you don't want to be visible while you are swiping pages.

#### Root cause
The issue arise with [6.2.0 version](https://github.com/callstack/react-native-pager-view/commit/769a3a21500fbe40c1909c9bf50f724c76a5b435#diff-b76afe0b22639c5e44b27b49c147284272978b94c8975b572b7a10cf8a3f5ba6R18) and support for navigate back gesture.

#### Solution
I've tested that navigate back gesture and it didn't worked for me as you may see on attached video.
I've also checked if basic react native ScrollView supports that for horizontal mode and it's not.
Based on all above I believe it's not worth to block users and I reverted that change.
I think we can revisit that use case later and fix that long living regression.

## Test Plan

| Architecture      | Before | After |
| ------- | :---------: | :---------: |
| Fabric     | <video src="https://github.com/user-attachments/assets/1051e996-402e-4cd9-a127-e7e475dba504"> | <video src="https://github.com/user-attachments/assets/17a54789-95fa-4ee2-8d42-4266cdba479c">  | 
| Paper |  <video src="https://github.com/user-attachments/assets/98e75c77-8268-4c5d-b3c3-baaad275ddce"> | <video src="https://github.com/user-attachments/assets/2b07c896-5490-4b66-af80-0dddf56bf91d"> |

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅    |
| Android |    ❌     |

## Checklist

### DIFF
--- a/example/src/tabView/NestedTabViewExample.tsx
+++ b/example/src/tabView/NestedTabViewExample.tsx
@@ -1,0 +1,55 @@
+import * as React from 'react';
+import { View, useWindowDimensions, Text, ScrollView } from 'react-native';
+import { TabView, SceneMap } from 'react-native-tab-view';
+import { Header } from 'react-native/Libraries/NewAppScreen';
+
+function FirstRoute() {
+  return (
+    <View style={{ flex: 1, padding: 20, backgroundColor: 'blue' }}>
+      <Text style={{color: 'white'}}>First Route</Text>
+    </View>
+  );
+}
+
+function SecondRoute() {
+  return (
+    <View style={{ flex: 1, padding: 20, backgroundColor: 'purple' }}>
+      <Text style={{color: 'white'}}>Second Route</Text>
+    </View>
+  );
+}
+
+const renderScene = SceneMap({
+  first: FirstRoute,
+  second: SecondRoute,
+});
+
+const routes = [
+  { key: 'first', title: 'First' },
+  { key: 'second', title: 'Second' },
+];
+
+export function NestedTabViewExample() {
+  const layout = useWindowDimensions();
+  const [index, setIndex] = React.useState(0);
+
+  return (
+    <ScrollView
+      contentContainerStyle={{flexGrow: 1, backgroundColor: 'red'}}
+      nestedScrollEnabled={false}
+      scrollEnabled={true}
+    >
+      <Header />
+
+      <View>
+        <TabView
+          style={{height: 1200}}
+          navigationState={{ index, routes }}
+          renderScene={renderScene}
+          onIndexChange={setIndex}
+          initialLayout={{ width: layout.width }}
+        />
+      </View>
+    </ScrollView>
+  );
+}
--- a/example/src/App.tsx
+++ b/example/src/App.tsx
@@ -36,8 +36,10 @@ import { createNativeStackNavigator } from '@react-navigation/native-stack';
 import { SafeAreaProvider } from 'react-native-safe-area-context';
 import { NextBasicPagerViewExample } from './NextBasicPagerViewExample';
 import { PagerHookExample } from './PagerHookExample';
+import { NestedTabViewExample } from './tabView/NestedTabViewExample';
 
 const examples = [
+  { component: NestedTabViewExample, name: 'Nested TabView Example' },
   { component: BasicPagerViewExample, name: 'Basic Example' },
   { component: PagerHookExample, name: 'Pager Hook Example' },
   { component: KeyboardExample, name: 'Keyboard Example' },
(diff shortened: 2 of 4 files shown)


MrRefactor:
This PR will be merged once we close this one: https://github.com/callstack/react-native-pager-view/pull/918

example/src/tabView/NestedTabViewExample.tsx
```diff
@@ -0,0 +1,56 @@
...
+    <ScrollView
+      contentContainerStyle={{flexGrow: 1, backgroundColor: 'red'}}
+      nestedScrollEnabled={false}
+      scrollEnabled={true}
+    >
+      <Header />
+
+      <View>
+        <TabView
+          // TODO: Fix the height measurement for nested pager view
```
MrRefactor, line 46: Please remove that TODO

MrRefactor (approved):

MateWW:
> @MateWW Could you check this PR https://github.com/callstack/react-native-pager-view/pull/771/files if this is the same thing?

yup, just in my case we have additional example in app to test that scenario

troZee:
> > @MateWW Could you check this PR https://github.com/callstack/react-native-pager-view/pull/771/files if this is the same thing?
> 
> yup, just in my case we have additional example in app to test that scenario

Thank you 🙏 we will check that change once we test other PRs, which are already in a queue

PR #941, opened 2024-12-02
fix(paper): prevent freeze on layout update during transition

krozniata:
# Summary
This PR fixes issue causing component to freeze when layout was updated during transition - old arch

When layout was updated e.g. because of hiding keyboard, setViewControllers completion block was never called, which resulted in swipe being disabled

It fixes this issue by adding custom shadow view and preventing layout update during transition phase

## Test Plan

1. Build the example app.
2. Go to the Keyboard Example menu.
3. Touch the TextInput to show the keyboard.
4. Press the Next page button. (might be required to modify ControlPanel to move button above keyboard)
5. Should navigate correctly, should be able to interact

Other examples should work as before

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    ❌     |

### DIFF
--- a/ios/RNCPagerViewManager.m
+++ b/ios/RNCPagerViewManager.m
@@ -1,5 +1,6 @@
 
 #import "RNCPagerViewManager.h"
+#import "RNCPagerViewShadowView.h"
 
 @implementation RNCPagerViewManager
 
@@ -30,7 +31,7 @@ RCT_EXPORT_VIEW_PROPERTY(layoutDirection, NSString)
             RCTLogError(@"Cannot find RNCPagerView with tag #%@", reactTag);
             return;
         }
-        if (!animated || !view.animating) {
+        if (!animated || !view.transitioning) {
             [view goTo:index.integerValue animated:animated];
         }
     }];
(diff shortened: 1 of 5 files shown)


ios/RNCPagerViewManager.m
```diff
@@ -1,5 +1,6 @@
 
 #import "RNCPagerViewManager.h"
+#import "RNCPagerViewShadowView.h"
```
troZee, line 3: is it still needed?

troZee (approved):

MrRefactor (approved):

ios/RNCPagerViewShadowView.m
```diff
@@ -0,0 +1,22 @@
...
+}
+
+- (void)layoutWithMetrics:(RCTLayoutMetrics)layoutMetrics
+            layoutContext:(RCTLayoutContext)layoutContext {
+    if (_transitioning) {
+      return;
+    }
+  
+    [super layoutWithMetrics:layoutMetrics layoutContext:layoutContext];
+}
```
okwasniewski, line 14: A small explanation why we don't want to trigger layout would be useful

okwasniewski (approved):

sueasueasuea:
Any update on this issue ?

floydkim:
hi, @krozniata @MrRefactor 
It’s been quite a while since the PR was approved.
Is there anything blocking the merge and release?

Issue #960, opened 2025-01-14
Swipes incorrectly trigger taps on Android if JS thread is busy


gaearon:
## Environment

- React Native 0.76.3 (but previous versions too)
- Reanimated 3.17 alpha (but previous versions too)
- React Native Pager View 6.6.1 (but previous versions too)
- Paper (have not checked Fabric yet)

## Description

If I nest `Pressable` into `PagerView`, and if I have a Reanimated event listener on it that makes the JS thread busy right after the swipe begins, the RN `Pressable` mistakingly recognizes a pager swipe as a press. I don't know exactly where the issue is but I suspect it should be up to this library to somehow prevent this from happening since it's the one emitting the scroll state change event (and thus should be able to disable/cancel touches inside in time).

This might be related to https://github.com/callstack/react-native-pager-view/issues/954 or might even be the same issue.

## Reproducible Demo

Here you go:

https://github.com/user-attachments/assets/3544c321-1855-4751-bd06-906679f9d95d

```js
import {Alert, View, Pressable} from 'react-native'
import Animated, {useEvent, useHandler, runOnJS} from 'react-native-reanimated'
import PagerView from 'react-native-pager-view'

const AnimatedPagerView = Animated.createAnimatedComponent(PagerView)

function doNothingSlowly() {
  let now = performance.now()
  while (performance.now() - now < 200) {
    // do nothing
  }
}

export default function App() {
  const handlePageScroll = useEvent(
    () => {
      'worklet'
      runOnJS(doNothingSlowly)()
    },
    ['onPageScrollStateChanged'],
    true,
  )
  return (
    <AnimatedPagerView style={{flex: 1}} onPageScroll={handlePageScroll}>
      <Page />
      <Page />
    </AnimatedPagerView>
  )
}

function Page() {
  return (
    <Pressable
      style={{
        width: 300,
        height: 200,
        backgroundColor: 'yellow',
        margin: 50,
      }}
      onPress={() => {
        Alert.alert('hello')
      }}
    />
  )
}
```

This reliably reproduces for me both on the simulator and on the device. If it doesn't for you, try tweaking the sleep ms number.

gaearon:
Application-level hackfix to work around this: https://github.com/bluesky-social/social-app/pull/7448

I'm not happy with it ofc but it may point to some way to solve it.

chrfalch:
After spending some time investigating this issue I think I have an idea of what's going on in your example. 

[...]

Issue #954, opened 2025-01-07
Android: Children don't properly cancel a press when pager starts scrolling


gpp-0:
## Environment

react-native: 0.76.5
react-native-pager-view: 6.6.1
Tested on Pixel 8

## Description

I believe that when the user swipes to scroll the pager and the pager takes control of the gesture, any children of the pager that were listening to the gesture should be sent a touchcancel event, and they should stop receiving subsequent events. This is for example how it works with FlatList.
Currently PagerView does not work like this. When a child component is pressed in and then a pager scroll is performed, the child receives a press out event but it never receives a cancel event, and instead it keeps listening to touch events, which leads to faulty behavior.
I have confirmed that in a native Android app ViewPager2 works as expected, an ACTION_CANCEL event is sent to the child, and the pager takes ownership of the gesture when a scroll begins.

## Reproducible Demo

Here is a snack demo, although this is easier to see on a real device
https://snack.expo.dev/@the_cetus/pagerview-touch?platform=android

### Steps to reproduce:
* Open your logs
* Staying on the first page, scroll up and down the list of buttons. You should see that whatever button you started the drag from receives a few touch events, but when you drag far enough eventually it receives a cancel event and it stops responding to further events, as expected.
* Now scroll horizontally to engage the pager. Keeping your finger down, you will see that the button keeps receiving move events indefinitely, even after it's received a press out and it looks like it should be ignoring events.
* Now for an actually visible bug, it's possible to do a very fast, very brief horizontal swipe in such a way that the pager scrolls to the next page, and a button also gets pressed. This is more easily done on a real device, but it requires precise timing so it doesn't happen often.

https://github.com/user-attachments/assets/1f19fab7-aff4-470a-8e88-a53221ec26d7

Here is what a typical log looks like when this happens
```
{"timestamp": 55806165, "type": "PRESS IN", "x": 115.992, "y": 105.986}
{"timestamp": 55806165, "type": "START", "x": 115.992, "y": 105.986}
{"timestamp": 55806175, "type": "MOVE", "x": 129.065, "y": 105.986}
{"timestamp": 55806206, "type": "MOVE", "x": 161.656, "y": 105.986}
{"timestamp": 55806206, "type": "PRESSED", "x": 148.989, "y": 105.986}
{"timestamp": 55806206, "type": "END", "x": 148.989, "y": 105.986}
{"timestamp": 55806206, "type": "PRESS OUT", "x": 148.989, "y": 105.986}
```
[...]

PR #961, opened 2025-01-16
fix(android): prevent taps while scrolling

gpp-0:
# Summary

This PR fixes a bug on Android where taps sometimes incorrectly trigger while swiping between pager screens, see issues https://github.com/callstack/react-native-pager-view/issues/960 and https://github.com/callstack/react-native-pager-view/issues/954. These issues in turn were opened in order to investigate https://github.com/bluesky-social/social-app/pull/7312
The cause of the bug seems to have had to do with React not being notified when the native pager starts scrolling, which meant that it didn't properly cancel gestures in the pager's children.

The fix involves adding a call to [notifyNativeGestureStarted](https://github.com/facebook/react-native/blob/c65117eba9f16da023e99253bfe28c0858f332fd/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/uimanager/events/NativeGestureUtil.kt), whose purpose is exactly to notify JS that a native gesture has started so it can dispatch the appropriate cancel events, see its use in other similar components like [ReactScrollView](https://github.com/facebook/react-native/blob/c65117eba9f16da023e99253bfe28c0858f332fd/packages/react-native/ReactAndroid/src/main/java/com/facebook/react/views/scroll/ReactScrollView.java#L454)

This call was added to NestedScrollableHost, where scrolling canceling logic was already present, however the behavior of nested scrollables should hopefully remain unchanged.

## Test Plan

I have tested the issues this PR aims to solve (https://github.com/callstack/react-native-pager-view/issues/960, https://github.com/callstack/react-native-pager-view/issues/954, https://github.com/bluesky-social/social-app/pull/7312) and I cannot reproduce the described bugs anymore. 
I have also tested all the examples in the repo, and all seem to function as expected. In particular, examples with scrollables (eg. ScrollViewInsideExample) or nested pagers (eg. NestPagerView) seem to be unaffected. 

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ❌     |
| Android |    ✅     |

## Checklist

### DIFF
--- a/android/src/main/java/com/reactnativepagerview/NestedScrollableHost.kt
+++ b/android/src/main/java/com/reactnativepagerview/NestedScrollableHost.kt
@@ -8,6 +8,7 @@ import android.view.ViewConfiguration
 import android.widget.FrameLayout
 import androidx.viewpager2.widget.ViewPager2
 import androidx.viewpager2.widget.ViewPager2.ORIENTATION_HORIZONTAL
+import com.facebook.react.uimanager.events.NativeGestureUtil
 import kotlin.math.absoluteValue
 import kotlin.math.sign
 
@@ -57,17 +58,14 @@ class NestedScrollableHost : FrameLayout {
   }
 
   private fun handleInterceptTouchEvent(e: MotionEvent) {
-    val orientation = parentViewPager?.orientation ?: return
-
-    // Early return if child can't scroll in same direction as parent
-    if (!canChildScroll(orientation, -1f) && !canChildScroll(orientation, 1f)) {
-      return
-    }
+    val orientation = parentViewPager?.orientation
 
     if (e.action == MotionEvent.ACTION_DOWN) {
       initialX = e.x
       initialY = e.y
-      parent.requestDisallowInterceptTouchEvent(true)
+      if (orientation != null) {
+        parent.requestDisallowInterceptTouchEvent(true)
+      }
     } else if (e.action == MotionEvent.ACTION_MOVE) {
       val dx = e.x - initialX
       val dy = e.y - initialY
@@ -78,6 +76,9 @@ class NestedScrollableHost : FrameLayout {
       val scaledDy = dy.absoluteValue * if (isVpHorizontal) 1f else .5f
 
       if (scaledDx > touchSlop || scaledDy > touchSlop) {
+        NativeGestureUtil.notifyNativeGestureStarted(this, e)
+
+        if (orientation == null) return
         if (isVpHorizontal == (scaledDy > scaledDx)) {
           // Gesture is perpendicular, allow all parents to intercept
           parent.requestDisallowInterceptTouchEvent(false)
--- a/android/src/main/java/com/reactnativepagerview/NestedScrollableHost.kt
+++ b/android/src/main/java/com/reactnativepagerview/NestedScrollableHost.kt
@@ -28,6 +28,7 @@ class NestedScrollableHost : FrameLayout {
   private var touchSlop = 0
   private var initialX = 0f
   private var initialY = 0f
+  private var nativeGestureStarted: Boolean = false
   private val parentViewPager: ViewPager2?
     get() {
       var v: View? = parent as? View
@@ -77,6 +78,7 @@ class NestedScrollableHost : FrameLayout {
 
       if (scaledDx > touchSlop || scaledDy > touchSlop) {
         NativeGestureUtil.notifyNativeGestureStarted(this, e)
+        nativeGestureStarted = true
 
         if (orientation == null) return
         if (isVpHorizontal == (scaledDy > scaledDx)) {
@@ -94,5 +96,15 @@ class NestedScrollableHost : FrameLayout {
         }
       }
     }
+  }
+
+  override fun onTouchEvent(e: MotionEvent): Boolean {
+    if (e.actionMasked == MotionEvent.ACTION_UP) {
+      if (nativeGestureStarted) {
+        NativeGestureUtil.notifyNativeGestureEnded(this, e)
+        nativeGestureStarted = false
+      }
+    }
+    return super.onTouchEvent(e)
   }
 }


gaearon:
Verified this fixes the issue in our app.

android/src/main/java/com/reactnativepagerview/NestedScrollableHost.kt
```diff
@@ -78,6 +76,9 @@ class NestedScrollableHost : FrameLayout {
       val scaledDy = dy.absoluteValue * if (isVpHorizontal) 1f else .5f
 
       if (scaledDx > touchSlop || scaledDy > touchSlop) {
+        NativeGestureUtil.notifyNativeGestureStarted(this, e)
```
gaearon, line 79: Does this need a symmetrical `notifyNativeGestureEnded` somewhere?

gaearon: Perhaps:

```diff
-@@ -57,17 +58,14 @@ class NestedScrollableHost : FrameLayout {
+@@ -27,6 +28,7 @@ class NestedScrollableHost : FrameLayout {
+   private var touchSlop = 0
+   private var initialX = 0f
+   private var initialY = 0f
++  private var nativeGestureStarted: Boolean = false
+   private val parentViewPager: ViewPager2?
+     get() {
+       var v: View? = parent as? View
+@@ -57,17 +59,14 @@ class NestedScrollableHost : FrameLayout {
    }
  
    private fun handleInterceptTouchEvent(e: MotionEvent) {
@@ -36,13 +44,26 @@ index 91d9946..294a3a8 100644
      } else if (e.action == MotionEvent.ACTION_MOVE) {
        val dx = e.x - initialX
        val dy = e.y - initialY
-@@ -78,6 +76,9 @@ class NestedScrollableHost : FrameLayout {
+@@ -78,6 +77,10 @@ class NestedScrollableHost : FrameLayout {
        val scaledDy = dy.absoluteValue * if (isVpHorizontal) 1f else .5f
  
        if (scaledDx > touchSlop || scaledDy > touchSlop) {
++        nativeGestureStarted = true
 +        NativeGestureUtil.notifyNativeGestureStarted(this, e)
 +
 +        if (orientation == null) return
          if (isVpHorizontal == (scaledDy > scaledDx)) {
            // Gesture is perpendicular, allow all parents to intercept
            parent.requestDisallowInterceptTouchEvent(false)
+@@ -92,6 +95,11 @@ class NestedScrollableHost : FrameLayout {
+           }
+         }
+       }
++    } else if (e.action == MotionEvent.ACTION_UP) {
++      if (nativeGestureStarted) {
++        nativeGestureStarted = false
++        NativeGestureUtil.notifyNativeGestureEnded(this, e)
++      }
+     }
+   }
+ }
```

gaearon: Hmm actually looking at RN code, `notifyNativeGestureEnded` is typically called from `onTouchEvent` instead

gpp-0: Lmao that's literally what I wrote on my end just now, like word for word. Will test it now, although I'm not sure what to test for because I don't know what NOT including this could break

gpp-0: Yeah, it definitely makes sense to call notifyNativeGestureEnded somewhere but a) i'm not sure what NOT including it will break so i don't know what to test for, and b) we are already sort of doing this touch event stuff in an unorthodox way so trying to follow other React code is confusing

gaearon: I think https://github.com/bluesky-social/social-app/pull/7459/commits/88b187815d95e5e3f76e61af22e76e5715751236 is the closest to what RN views do

gaearon: I figured following these should be a safe pattern: https://github.com/search?q=repo%3Afacebook%2Freact-native+notifyNativeGestureEnded&type=code

Agree it sucks we don't know what to test for. But every `notifyNativeGestureStarted` in the RN codebase follows the same pattern:

- It happens in `onInterceptTouchEvent`
- It sets an instance variable
- That variable gets reset in `onTouchEvent` along with `notifyNativeGestureEnded`

There are no cases with a different pattern, and there are no cases where `*Started` is missing an `*Ended` pair so I think even in the absence of extra testing cases, following this makes sense.

gpp-0: Yep that makes sense, I'll add it here too! By the way, overriding onTouchEvent raises a [ClickableViewAccessibility](https://googlesamples.github.io/android-custom-lint-rules/checks/ClickableViewAccessibility.md.html) warning, although I don't see this addressed in any of the React code, is it a problem for us?

gaearon: I assume not because we're not running any app logic and are just coordinating things here. As opposed to, say, actually handling clicks or something.

gpp-0: Well apparently it's fine because `onTouchEvent` NEVER gets called anyway? In general `onTouchEvent` only gets called after `onInterceptTouchEvent` has returned true, but in this case we just call `super.onInterceptTouchEvent` which does something completely irrelevant and always returns false. Which it's definitely not supposed to do. And yet it works. Honestly this class might be in need of a complete overhaul. Hopefully someone else can handle that :)

gaearon: lol

troZee:
Hello 👋 
Thank you for your 1st contribution to the pager library. In this PR, I fixed the CI https://github.com/callstack/react-native-pager-view/pull/962 . Could you rebase your PR with the main branch? 

Once it is rebased, we will merge it 

Thank you for your cooperation 🙏

troZee (approved):

MrRefactor (approved):
Looks good to me!

troZee:
@gpp-0 @gaearon 
https://github.com/callstack/react-native-pager-view/releases/tag/v6.7.0

gaearon:
fantastic, thank you!

Issue #946, opened 2024-12-10
[Android] App crashes on navigating out of the screen with Pager


MrRefactor:
## Environment
Android, latest version 6.6.1

## Description
App crashes on navigating out of the screen with Pager, checking the rootcause

Jahanzeb009:
don't use `pageMargin`, it solves my problem on RN 0.76.5

JoniVR:
@MrRefactor
Is this also throwing the same exception as #859 for you? 

Because I'm still facing this (even on 6.6.1) but I also managed to find a fix that seems to resolve it and is stable. I just haven't submitted a PR yet since I seem to be the only one facing this crash based on the issues.

MrRefactor:
Resolved in 6.8.0 version

PR #986, opened 2025-04-08
fix: Try to ignore IllegalArgumentException

NiuGuohui:
Try to ignore IllegalArgumentException. This issue: #946

I don't think this error is a fatal one, so it shouldn't cause the app to crash.


MrRefactor:
Closed as per https://github.com/callstack/react-native-pager-view/pull/1001

Issue #859, opened 2024-08-20
App is getting crashed when switching the navigator, java.lang.IllegalArgumentException: Scrapped or attached views may not be recycled. isScrap:false isAttached:true


rifad4u:
## Environment

"@react-navigation/material-top-tabs": "^6.6.14",
"@react-navigation/native": "^6.1.18",
"@react-navigation/native-stack": "^6.11.0",
"react": "18.3.1",
"react-native": "0.75.2",
"react-native-pager-view": "^6.3.4",
"react-native-safe-area-context": "^4.10.9",
"react-native-screens": "^3.34.0",
"react-native-tab-view": "^3.5.2"

## Description

I have a tab navigator in a post login stack screen, one of the tabs visibility is based on a value of state variable. so when i do logout first i'm setting the value of the state variable to false, then i'm switching the stack to pre login stack. at this time app is getting crashed

Getting logs in Android studio like below
java.lang.IllegalArgumentException: Scrapped or attached views may not be recycled. isScrap:false isAttached:true androidx.viewpager2.widget.ViewPager2$RecyclerViewImpl{7d03f4b VFED..... .......D 0,0-1080,1736 #2}, adapter:com.reactnativepagerview.ViewPagerAdapter@5f5fd28, layout:androidx.viewpager2.widget.ViewPager2$LinearLayoutManagerImpl@1cf2a41, context:com.facebook.react.uimanager.ThemedReactContext@c1e63c

## Reproducible Demo

https://github.com/rifad4u/pagerviewcrash.git

jingp-ofload:
/+ 1

rifad4u:
@MrRefactor @yilinjuang  @WoLewicki  @okwasniewski  @szymonrybczak @szydlovsky  any update on this issue?

troZee:
Hi @rifad4u

Just a quick reminder to please avoid pinging other contributors unnecessarily. Let's ensure we're not disturbing others unless it's critical to the discussion. Thanks for your understanding and cooperation! PRs and other investigations are welcome 🙏 

Best regards,
Piotr Trocki

troZee:
@kamill90

ShaoGongBra:
The same problem

sankar-gp:
Any update on this issue?

sankar-gp:
Im facing this issue on react native 0.75.4 with "react-native-pager-view": "^6.4.1",

LinhLM23496:
+ 1

[...]

PR #1001, opened 2025-05-20
fix(android): fix crash on pager unmount

### LINKED ISSUE #946: [Android] App crashes on navigating out of the screen with Pager

krozniata:
# Summary

This PR fixes issue with crashing on android when leaving screen using native stack

Root cause of that crash was that component tried to call queued `refreshViewChildrenLayout` on unmounted views - when leaving screen all of pages were firstly being unmounted and then `refreshViewChildrenLayout` method was called

Fixes #946 #859 

## Test Plan

1. Open example app on android with native stack
2. Go to any example
3. Go back to example list
4. App shouldn't crash


MrRefactor (approved):
lgtm, will include that in next release

PR #1074, opened 2026-04-07
fix(ios): guarantee completion block fires and fix swipe permanently disabled after out-of-range goTo (paper arch)

Norcy:
## Summary

Two related bugs in the **paper (Old Architecture) iOS implementation** (`ios/RNCPagerView.m`), both causing the pager to become permanently unresponsive to swipe gestures:

### Bug 1 — swipe stuck after an out-of-range `goTo` call

`goTo:` always calls `disableSwipe` before the bounds check, but the early-return path for out-of-range indices never calls `enableSwipe`. If any component (e.g. a parent scroll view or animated list) triggers `goTo:` with an index that is briefly out of range, the pager locks up.

**Fix:** call `[self enableSwipe]` before the early `return`.

### Bug 2 — completion block silently dropped by UIPageViewController

`UIPageViewController -setViewControllers:direction:animated:completion:` is documented to always call its completion block, but in practice it **does not** when:
- the method is called while a previous transition is still in flight, or
- the view controller's view is not yet in the window hierarchy.

When the block is never called, `enableSwipe` is never invoked and `transitioning` stays `YES`, permanently locking the pager.

**Fix:** introduce `UIPageViewController+SafeSet` Objective-C category that wraps the call with:
- a 1-second watchdog timer that fires `completion(NO)` if the system callback never arrives,
- a debounce mechanism so rapid successive calls cancel the previous pending timer, and
- a `hasCompleted` one-shot guard to prevent double-invocation.

## Files changed

| File | Change |
|------|--------|
| `ios/RNCPagerView.m` | Add `enableSwipe` on early return; use `safeSetViewControllers:` |
| `ios/UIPageViewController+SafeSet.h` | New category declaration |
| `ios/UIPageViewController+SafeSet.m` | New category implementation |

## Context

This fix targets the paper-architecture code (`RNCPagerView.m`) which was removed in this repository after v6.8.1. The npm package `react-native-pager-view@6.9.1` (published from git commit `9c8446f`) still contains this code, so there are users in the wild who are affected. I'm submitting this PR to document the fix and to give maintainers the option of issuing a patch release for the v6.x line.

The same `setViewControllers` risk also exists in `ios/Fabric/RNCPagerViewComponentView.mm` (present through v7.x), though that is a separate scope.

## Test plan

- [ ] Navigate to a page index that is temporarily out of range while pages are loading; confirm swipe remains functional afterwards
- [ ] Rapidly call `scrollTo` / `setPage` in quick succession; confirm the pager does not freeze and `onPageSelected` fires for the final destination
- [ ] Verify no double-invocation of `onPageSelected` under normal navigation

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/ios/RNCPagerView.m
+++ b/ios/RNCPagerView.m
@@ -8,6 +8,7 @@
 #import "RCTOnPageScrollStateChanged.h"
 #import "RCTOnPageSelected.h"
 #import <math.h>
+#import "UIPageViewController+SafeSet.h"
 
 @interface RNCPagerView () <UIPageViewControllerDataSource, UIPageViewControllerDelegate, UIScrollViewDelegate>
 
@@ -178,10 +179,10 @@
       [self setTransitioning:YES];
     }
     
-    [self.reactPageViewController setViewControllers:@[controller]
-                                           direction:direction
-                                            animated:animated
-                                          completion:^(BOOL finished) {
+    [self.reactPageViewController safeSetViewControllers:@[controller]
+                                               direction:direction
+                                                animated:animated
+                                              completion:^(BOOL finished) {
         __strong typeof(self) strongSelf = weakSelf;
         strongSelf.currentIndex = index;
         strongSelf.currentView = controller.view;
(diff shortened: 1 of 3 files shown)


MrRefactor:
Hey, with 7.0.0 version we dropped the old arch support, due to that Im closing this PR.

PR #1078, opened 2026-05-04
fix(ios): persist position during page list head changes

### LINKED ISSUE #965: (iOS) Dynamic pages causes page switch when adding pages at the start
## Environment

Expo v52
React Native 0.76.6
Expo Router v4
iPhone 16 Pro Simulator

### Summary

I'm using this pager view to show images (similar to iOS photos app), and since there's a lot of images to be loaded, it's images are loaded as requested, similar to FlatList `onStartReached` and `onEndReached`.

### Current behaviour

Adding pages to the end of the data array works very well, but when adding to the start, the current page changes to the new page added based on the page index it was before.

### Expected behaviour

I can see this might be an intended behaviour based on what is being built, but having at least a prop like `preservePage` to update the current page while preserving the page in view (maybe based on page key or id) would be perfect for dynamically loading pages.

I also tested on Android, and it works as expect most of the times, maintaining the correct current page view. Some times it does not update the `activePage` correctly, but the current page view stay as is.

## Findings

On version `6.5.1` of this library, the current page view is preserved as expected, and after digging deeper, I found out that #923 (introduced on `6.6.0`) is the "culprit" for the issue. Although this makes the current page view still visible, the actual `activePage` is not updated accordingly, which also impacts on the position given by `onPageScroll`. It only correctly updates the `activePage` when completing a swipe to the left/right, then `onPageSelected` is called with the correct position.

alpha0010:
# Summary

Dynamic changes to the page list prior to the currently visible page now maintain the same visible page on iOS (this already works on Android).

Fixes #965

```ts
function DemoPager() {
  const allPages = [0, 1, 2, 3, 4];
  const [pages, setPages] = useState(allPages);

  const togglePage = (page: number) =>
    setPages((prev) => {
      const itemIndex = prev.indexOf(page);
      const next = [...prev];
      if (itemIndex === -1) {
        next.push(page);
        next.sort((a, b) => a - b);
      } else {
        next.splice(itemIndex, 1);
      }
      return next;
    });

  return (
    <View style={styles.container}>
      <PagerView style={styles.container}>
        {pages.map((page) => (
          <View collapsable={false} key={`x:${page}`}>
            <Text>{page}</Text>
          </View>
        ))}
      </PagerView>
      <View style={styles.container}>
        {allPages.map((page) => (
          <Button
            key={`x:${page}`}
            onPress={() => togglePage(page)}
            title={`Toggle ${page}`}
          />
        ))}
      </View>
    </View>
  );
}

const styles = StyleSheet.create({
  container: {flex: 1},
});
```

Use the above test screen to add/remove pages before/after the currently visible page.  With this PR, iOS should now match the existing Android behavior.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |    ✅     |
| Android |    N/A    |

## Checklist

### DIFF
--- a/ios/PagerViewProvider.swift
+++ b/ios/PagerViewProvider.swift
@@ -74,6 +74,9 @@ import UIKit
       return
     }
     props.children.insert(IdentifiablePlatformView(child), at: index)
+    if index <= props.currentPage {
+      props.currentPage += 1
+    }
   }
 
   @objc(removeChildAtIndex:)
@@ -82,6 +85,9 @@ import UIKit
       return
     }
     props.children.remove(at: index)
+    if index < props.currentPage {
+      props.currentPage -= 1
+    }
   }
 
   override public func didMoveToWindow() {
--- a/ios/PagerViewProvider.swift
+++ b/ios/PagerViewProvider.swift
@@ -73,10 +73,10 @@ import UIKit
     guard index >= 0 && index <= props.children.count else {
       return
     }
-    props.children.insert(IdentifiablePlatformView(child), at: index)
-    if index <= props.currentPage {
+    if index <= props.currentPage && props.currentPage < props.children.count {
       props.currentPage += 1
     }
+    props.children.insert(IdentifiablePlatformView(child), at: index)
   }
 
   @objc(removeChildAtIndex:)


alpha0010:
Discovered a potential issue with this implementation, investigating.

alpha0010:
So, I am at a bit of a loss.  After the following data transition:
```
Before:
A B C D E
    ^

After:
B X C Y D
    ^
```
Page "C" is still correctly the active page, but instead of displaying content, renders blank white.  The neighboring pages still render correctly, and if I swipe away and swipe back, "C" will render again.  But this error does not occur on initial load, only after using the pager some.

Any insights you might have would be greatly appreciated.  I feel like I must be missing some fundamental concept, since all my logging and debugging efforts seem to say everything is working as intended.

alpha0010:
Documenting my testing:
- Manage tabs/current page via `id` (from `RepresentableView`) instead of index
  - Seems to work essentially the same as the current implementation (including exhibiting the blank screen bug from the above described data transition)
- In `.onChange(of: props.children)`, if the above data transition is detected, increment the `TabView.id()` to force a rebuild
  - Fixes the initial blank problem
  - Navigating away and back programmatically works fine
  - Swiping away and back triggers the blank screen (the first time; the second time swiping away/back, it displays correctly)
- Most related I could find is: https://stackoverflow.com/questions/59450948/swiftui-why-is-tabview-and-its-contents-sometimes-not-refreshing-when-i-change , which makes me wonder if this logic path is hitting a bug in `TabView`'s internal implementation.

alpha0010:
What works:
- Correct active page is maintained when inserting/deleting at any index, on both iOS and Android
  - Note: this already was the state on Android, but now iOS works as well

What fails:
- On iOS, inserting a page immediately prior to the active page (while simultaneously dropping the list head page), swiping away, then swiping back, displays a blank screen (on next swipe, the screen correctly renders again)
  - Workaround in my app, is to pass the pager a wrapper array of `View`s, that manage actual insertions by swapping children, but as far as the pager knows, only insert/delete at the head/tail of the array
- On Android, consecutive animated programmatic page turns (with state changes to the children pages in between), result in the programmatic page turn navigating to the wrong index
  - Workaround in my app is to use non-animated programmatic page turns (since this does not trigger the bug)

---

I have been unable solve either of these bugs, but since they have workarounds, I cannot really justify spending more dev time on them currently.  Marking this as ready (we already use it locally via patches).  I think it is an improvement over this library's current status, despite not solving all the dynamic page list related bugs.

PR #1085, opened 2026-06-21
fix(ios): propagate safe area insets to child UIKit views inside PagerView

### LINKED ISSUE #1090: [iOS] New Arch: safe area insets are stripped from PagerView children
## Environment

| Package | Version |
| --- | --- |
| react-native-pager-view | 8.0.0 – 8.0.2 (current `master`) |
| react-native | 0.85.x |
| Architecture | New Architecture (Fabric), iOS |

## Description

On the New Architecture iOS path, content rendered **inside** a `PagerView` page does **not** receive the screen's safe area insets. UI that relies on the safe area (content near the notch / Dynamic Island / home indicator) ends up with **zero insets** and can be clipped or sit under system UI.

## Root cause

In `ios/PagerViewProvider.swift`, the SwiftUI host is created with `ignoreSafeArea: true`:

```swift
self.hostingController = UIHostingController(
  rootView: PagerView(props: props, delegate: delegate),
  ignoreSafeArea: true
)
```

That convenience initializer (in `ios/Extensions.swift`) calls `disableSafeArea()`, which dynamically subclasses the hosting controller's view to return `safeAreaInsets = .zero`. This makes the pager render edge-to-edge, but as a side effect the hosted UIKit / React Native **child** views also report `safeAreaInsets = .zero`, so any RN content inside a page loses the real safe area.

`RepresentableView` (the wrapper for each page) is a plain `UIViewRepresentable` that just adds the RN view into a wrapper `UIView`, so there is nothing to restore the insets the host stripped.

## Expected behavior

The pager can be edge-to-edge, but the child pages should still receive the correct safe area insets (matching a non-paged screen), so RN layouts using `useSafeAreaInsets()` / `SafeAreaView` behave the same inside and outside the pager.

## Proposed fix

- Use SwiftUI `.ignoresSafeArea()` on the `TabView` to keep the pager edge-to-edge.
- Render each child through a `UIViewControllerRepresentable` (`PageChildViewController`) that re-injects the hosting view's real safe area insets via `additionalSafeAreaInsets` in `viewDidLayoutSubviews()`.
- Drop `ignoreSafeArea: true` from `setupView()` since the controller-level safe-area disabling is no longer needed.

PR: #1085

IsaacIsrael:
Fixes #1090 #1096

## Problem

When using `react-native-pager-view` inside a screen with iOS safe area (e.g. behind a translucent tab bar on iOS 26 liquid glass), child UIKit views — including `UIScrollView` / `FlashList` — receive `safeAreaInsets = .zero`. This causes `contentInsetAdjustmentBehavior` to have no effect: scroll views cannot automatically adjust their content insets to account for overlapping system chrome.

### Root cause

The current implementation uses `UIViewRepresentable` to embed React Native views inside SwiftUI pages. The `UIHostingController` is initialized with `ignoreSafeArea: true`, which dynamically subclasses the hosting view to override `safeAreaInsets` → `.zero`. Even without `ignoreSafeArea`, SwiftUI's layout system does not propagate safe area insets to child UIKit views embedded via `UIViewRepresentable`.

**Debug evidence** (from native logging added during investigation):

```
HostingController.view.safeAreaInsets: bottom: 83     ← correct
PagingCollectionView.safeAreaInsets:   bottom: 83     ← correct
Wrapper UIView.safeAreaInsets:         bottom: 0      ← LOST
RCTEnhancedScrollView.safeAreaInsets:  bottom: 0      ← zero → contentInsetAdjustmentBehavior has nothing to adjust
```

## Solution

Three targeted changes:

1. **`Extensions.swift`** — Replace `UIViewRepresentable` with `UIViewControllerRepresentable`. The new `PageChildViewController` traverses up to `_UIHostingView`, reads its real `safeAreaInsets`, and re-injects them via `additionalSafeAreaInsets`. This restores proper safe area propagation to all descendant UIKit views.

2. **`PagerView.swift`** — Add `.ignoresSafeArea()` to the SwiftUI `TabView` body so page content extends visually behind system chrome (translucent tab bars, etc.), enabling the "scroll behind glass" effect.

3. **`PagerViewProvider.swift`** — Remove `ignoreSafeArea: true` from `UIHostingController` init so the hosting view preserves real device insets for `PageChildViewController` to read.

### How it works

```
HostingController.view (bottom: 83)  ← real insets preserved (no more ignoreSafeArea)
  → PagingCollectionView (bottom: 83)
    → [.ignoresSafeArea() — content extends behind tab bar]
      → PageChildViewController
        → additionalSafeAreaInsets = hosting.safeAreaInsets  ← RE-INJECTED
          → Child UIKit views (bottom: 83)  ← contentInsetAdjustmentBehavior works!
```

The guard with 0.5pt threshold prevents infinite layout loops when `viewDidLayoutSubviews` sets `additionalSafeAreaInsets`.

## Test plan

- [ ] Verify `contentInsetAdjustmentBehavior='scrollableAxes'` works on scroll views inside PagerView pages (last item should stop above the tab bar when scrolled to bottom)
- [ ] Verify content scrolls behind translucent tab bar (liquid glass effect on iOS 26)
- [ ] Verify no layout regression on non-liquid-glass screens (pre-iOS 26)
- [ ] Verify horizontal paging still works correctly
- [ ] Verify no infinite layout loops (check CPU usage)

Made with [Cursor](https://cursor.com)


troZee:
@IsaacIsrael 
Could you fix the conflicts?

IsaacIsrael:
@troZee  done :)

ios/Extensions.swift
```diff
@@ -4,19 +4,60 @@ import UIKit
...
+    propagateHostingSafeArea()
+  }
+
+  /// Traverses up to the _UIHostingView and re-applies its bottom safe area
+  /// as additionalSafeAreaInsets on this child VC, since .ignoresSafeArea()
+  /// causes UIKit to report safeAreaInsets = .zero for embedded views.
+  private func propagateHostingSafeArea() {
+    var current: UIView? = view.superview
+    while let v = current {
+      if String(describing: type(of: v)).contains("_UIHostingView") {
```
troZee: Could you add an else statement and log this error if _UIHostingView is not found?

troZee:
@IsaacIsrael 
I added a different code, but I was not able to test it. Could you check this code and provide a repro code? 

In my PR, I applied the safer safe-area propagation in ios/Extensions.swift.

  What changed:
  - Removed the private _UIHostingView class-name lookup.
  - Added viewSafeAreaInsetsDidChange() so updates happen when UIKit safe-area values change.
  - propagateSafeArea() now uses the nearest UIKit superview with non-zero safeAreaInsets, falling back to
    view.window?.safeAreaInsets.

troZee:
@wonhyunkwon Could you test this PR?

wonhyunkwon:
@troZee 
Tested this PR against our production app and it fixes the exact issue I reported in [#1097](https://github.com/callstack/react-native-pager-view/pull/1097). 👍

How I applied it
Applied the changes on top of the 8.0.0 release via a pnpm patch. Note the PagerViewProvider.swift hunk didn't apply as-is since the guard let parentViewController refactor only exists on main — on 8.0.0 the equivalent change is just removing the ignoreSafeArea: true argument from the UIHostingController init. Extensions.swift and PagerView.swift applied cleanly.

Environment
- react-native-pager-view 8.0.0 + this PR's changes
- React Native 0.83.4 (new architecture), Expo SDK 55, @gorhom/bottom-sheet 5.2.6
- Physical iPhone 15 Pro (iOS 26.5) — this bug only reproduces on a physical device; the simulator's SwiftUI keyboard avoidance behaves differently, so simulator testing won't catch it

Scenario
PagerView rendered inside a @gorhom/bottom-sheet with a BottomSheetTextInput above it.
- Before: focusing the input triggers SwiftUI TabView's automatic keyboard avoidance, shrinking the pager and leaving a blank gap above the keyboard (RN layout values stay correct — it's purely native-level)
- After: the pager keeps its RN-driven layout when the keyboard appears — no gap ✅

Also checked our other PagerView usages (comment bottom sheet) and saw no safe-area regressions from the .ignoresSafeArea() + inset re-injection approach.

As for #1097, I'll leave it open for now and defer to the maintainers on whether to close it in favor of this PR — feel free to close it if this one moves forward.

troZee:
https://github.com/callstack/react-native-pager-view/releases/tag/v8.0.4

IsaacIsrael:
Sorry @troZee , I missed your message and thanks @wonhyunkwon for testing :)

troZee:
@IsaacIsrael could you test the newest version of it works for you? 🙏

iffa:
After this change it seems like expo router's js-top-tabs (which uses this lib under the hood) has some weirdness going on with the top insets: content goes under tab bar. looks to only happen on ios 26+ (i can't reproduce on ios 18 or android).

novatech:
> After this change it seems like expo router's js-top-tabs (which uses this lib under the hood) has some weirdness going on with the top insets: content goes under tab bar. looks to only happen on ios 26+ (i can't reproduce on ios 18 or android).

face same behavior with material-top-tab-navigator. need to wrap my component with
```
<SafeAreaView edges={[]}>
<MyComponent/>
</SafeAreaView>
```

currently reverting to 8.0.2 and the issue is not there

Issue #1056, opened 2026-01-28
[iOS] Interactive keyboardDismissMode shows a white space in place of keyboard


aminerol:
## Environment

- The example app on master branch 

## Description

Scrolling on list with `keyboardDismissMode` set to `interactive` seems to render a white view that goes under keyboard and it stays on as the keyboard is being dismissed, it only goes away after keyboard fully dismissed

## Reproducible Recording

https://github.com/user-attachments/assets/f98c56cc-723b-4883-b557-dbce88fb4f83

## Reproducible Demo

Its reproducible with example app by adding keyboardDismissMode prop to this [Chat's](https://github.com/callstack/react-native-pager-view/blob/master/example/src/tabView/Shared/Chat.tsx) component set to interactive

PR #1097, opened 2026-07-09
fix(ios): prevent SwiftUI keyboard avoidance from shrinking pages

### LINKED ISSUE #1096: [iOS] v8 (SwiftUI TabView): pages shrink when the keyboard appears — real devices only
## Environment

- react-native-pager-view: 8.0.0 (reproduced on 8.0.3 sources; the relevant code is unchanged up to current `main`)
- react-native: 0.83.4 (new architecture), Expo SDK 55
- iOS: reproduced on iPhone 15 Pro (iOS 26.x) — **real devices only**, NOT reproducible on simulators (tested iOS 18.6 and 26.1 simulators)

## Description

Since v8 the iOS implementation is a SwiftUI `TabView`. SwiftUI applies **automatic keyboard avoidance** by default: when the keyboard appears, the keyboard safe area shrinks the `TabView`'s content — so every page is resized **natively, outside of React Native's layout**.

In a React Native app, layout is fully owned by Yoga/Fabric, and keyboard handling is typically done at the RN level (`KeyboardAvoidingView`, `@gorhom/bottom-sheet` keyboard behaviors, etc.). The result is a **double adjustment**: the RN side already compensates for the keyboard, then SwiftUI shrinks the pages again, leaving a blank strip above the keyboard where page content should be.

`Extensions.swift` already tries to neutralize safe area influence for the hosting controller (`disableSafeArea()` overrides `safeAreaInsets` to `.zero`), but SwiftUI's keyboard avoidance is driven by the keyboard safe area channel, which that override does not cover.

## Reproduction

1. Place a `PagerView` inside any screen where the keyboard can open while the pager is visible (e.g. a bottom sheet with a search `TextInput` above the pager).
2. On a **real device**, focus the input.
3. The pager's pages shrink by roughly the keyboard height even though the RN layout of the pager itself did not change → blank area above the keyboard.

Diagnostics that pinpointed this:
- All RN layout values (`onLayout` of the pager and its children) stay **correct** the whole time — only the native frames shrink.
- Painting each layer a distinct background color shows the blank strip is the `PagerView`'s own background showing through where the (natively shrunken) pages should be.
- Why simulators don't reproduce it: SwiftUI keyboard avoidance behaves differently with the simulator's hardware-keyboard integration.

## Fix

One line — opt the `TabView` out of keyboard avoidance, which is the correct semantics for an RN-managed view:

```swift
.tabViewStyle(.page(indexDisplayMode: .never))
.ignoresSafeArea(.keyboard) // layout is fully driven by React Native
```

Verified on a real device (iPhone 15 Pro): pages keep their RN-driven size with the keyboard open, and no regression on simulators.

I'll open a PR with this change right away.

Possibly related: #1056 (white area during interactive keyboard dismiss) may share this root cause.

wonhyunkwon:
# Summary

Fixes #1096. Possibly also addresses #1056.

Since v8 the iOS implementation is a SwiftUI `TabView`, which applies **automatic keyboard avoidance** by default: when the keyboard appears, the keyboard safe area shrinks the `TabView`'s content, so every page is resized **natively, outside of React Native's layout**.

In a React Native app layout is fully owned by Yoga/Fabric, and keyboard handling is done at the RN level (`KeyboardAvoidingView`, `@gorhom/bottom-sheet` keyboard behaviors, etc.). The result is a double adjustment: the RN side already compensates for the keyboard, then SwiftUI shrinks the pages again, leaving a blank strip above the keyboard where page content should be. All RN layout values (`onLayout`) stay correct the whole time — only the native frames shrink, which makes this very hard to debug from the JS side.

`Extensions.swift` already neutralizes safe area influence for the hosting controller (`disableSafeArea()` overrides `safeAreaInsets` to `.zero`), but SwiftUI's keyboard avoidance is driven by the keyboard safe area channel, which that override does not cover.

**The fix is one line**: opt the `TabView` out of keyboard avoidance with `.ignoresSafeArea(.keyboard)`. This restores the correct semantics for an RN-managed view — native views should follow the RN-driven frame only.

Note: this only reproduces on **real devices**; simulators behave differently with SwiftUI keyboard avoidance (hardware-keyboard integration), which is likely why it slipped through.

- A **real iOS device** (reproduction is not possible on simulators)
- Any screen where the keyboard can open while a `PagerView` is visible — e.g. a `TextInput` above a `PagerView` (a bottom sheet with a search input works well)

1. Open the screen containing the `PagerView`.
2. Focus the `TextInput` so the keyboard appears.
3. **Before this fix**: the pages shrink by roughly the keyboard height even though the pager's RN layout is unchanged — a blank strip appears above the keyboard (giving each layer a distinct `backgroundColor` shows the strip is the `PagerView`'s own background where the natively-shrunken pages should be).
4. **After this fix**: pages keep their RN-driven size; content fills right up to the keyboard.

Verified in a production app (RN 0.83.4, new architecture):
- iPhone 15 Pro (real device): blank strip gone, pages keep RN-driven size with the keyboard open
- iPhone 16 Plus simulator (iOS 18.6) and iPhone 16 Plus simulator (iOS 26.1): no regression — behavior unchanged

`bun typescript`, `bun lint`, and `bun test` all pass (native-only change).

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     |     ✅      |
| Android |     ❌      |

## Checklist

- [ ] I added the documentation in `README.md` (not applicable — no API change)
- [ ] I updated the typed files (TS and Flow) (not applicable — native-only change)

### DIFF
--- a/ios/PagerView.swift
+++ b/ios/PagerView.swift
@@ -18,6 +18,9 @@ struct PagerView: View {
     .id(props.children.count)
     .background(.clear)
     .tabViewStyle(.page(indexDisplayMode: .never))
+    // Opt out of SwiftUI's automatic keyboard avoidance: layout is fully driven
+    // by React Native, so pages must not shrink when the keyboard appears.
+    .ignoresSafeArea(.keyboard)
     .environment(\.layoutDirection, props.layoutDirection.converted)
     .introspect(.tabView(style: .page), on: .iOS(.v14...)) { collectionView in
       self.collectionView = collectionView


troZee:
This PR is probably duplicated by this https://github.com/callstack/react-native-pager-view/pull/1085/changes

troZee:
Closed in favour of https://github.com/callstack/react-native-pager-view/pull/1085

PR #1105, opened 2026-08-27
fix(android): give ComposeView its own Lifecycle to fix #1103 and #1104

### LINKED ISSUE #1103: [Android] Tab content stays blank after navigating away and back over a PagerView (regression 8.0.5 → 9.0.2)
## Environment
                                                                                                                                                                                                        
  - react-native-pager-view: 9.0.2 (last working version: 8.0.5)                                                                                                                                                           
  - @react-navigation/material-top-tabs: 7.6.15 (react-native-tab-view 4.3.2)                                                                                                                                              
  - react-native: 0.86.2                                                                                                                                                                                                   
  - react: 19.2.3                                                                                                                                                                                                          
  - Platform: Android                                                                                                                                                                                                      
  - New architecture: enabled

## Description

A screen using createMaterialTopTabNavigator (even with a single tab, so a single-page PagerView) goes blank after the user navigates to another screen pushed on top via a Stack.Navigator, then navigates back. The tab bar and navigation state are still correct, but the page content is empty — no swipe/resize recovers it.                                                                                                                 
                                                                                                                                                                                                                           
Bisected on our end: reverting only react-native-pager-view from 9.0.2 to 8.0.5 (no other change) fixes it, so this is a regression introduced in 9.0.0. That release replaced the Android implementation from ViewPager2/RecyclerView with a Jetpack Compose HorizontalPager (ComposePagerView, #1092). Suspicion: the Compose-hosted page doesn't get a fresh re-composition/re-layout when the surrounding Fragment/Activity view is covered by another screen and then revealed again. 

## Reproducible Demo
  1. Screen A is a createMaterialTopTabNavigator with a single tab rendering some content (a PagerView under the hood, even with just 1 page).                                                                             
  2. From that tab's content, navigate forward to a screen pushed on a Stack.Navigator on top of it.                                                                                                                       
  3. Navigate back (pop the stack screen) to Screen A.                                                                                                                                                                     
  4. The tab's content area is blank — the PagerView page no longer renders anything, even though the tab bar/navigation state is correct. No user interaction (swipe/resize) recovers it.

### LINKED ISSUE #1104: [Android] PagerView 9 remounts/recreates page after navigating back to Material Top Tabs
# [Android] PagerView 9 remounts/recreates page after navigating back to Material Top Tabs

## Description

After upgrading `react-native-pager-view` from `8.0.2` to `9.x`, a page inside `@react-navigation/material-top-tabs` appears to be recreated/remounted when navigating to another screen and then going back.

This causes components such as `FlatList` inside the tab to be rebuilt and lose their previous UI state.

The issue only occurs with PagerView 9.x on Android.

Downgrading to `react-native-pager-view@8.0.2` fixes the issue.

## Steps to reproduce

1. Create a Material Top Tab Navigator.
2. Render a `FlatList` inside one of the tabs.
3. Navigate from the Top Tab screen to another Stack screen.
4. Press Back to return to the Top Tab screen.
5. Observe that the tab/page is recreated and the `FlatList` is rebuilt.

Example navigation flow:

```text
Stack
 ├── TopTabs
 │    ├── Tab A → FlatList
 │    └── Tab B
 │
 └── Detail

Tab A
  → Detail
  → Back
  → Tab A / FlatList rebuilt
```

### react-native-pager-view 9.x

```text
Open Tab A
→ Navigate to Detail
→ Back
→ Page/scene recreated
→ FlatList rebuilt
```

Issue occurs.

### react-native-pager-view 8.0.2

```text
Open Tab A
→ Navigate to Detail
→ Back
→ Existing TopTab/page state is preserved
```

Issue does not occur.

The application code and navigation configuration are unchanged between these two tests. Only the `react-native-pager-view` version is changed.

## Expected behavior

Returning from the Detail screen should preserve the existing PagerView page/scene, as it did with `react-native-pager-view@8.0.2`.

The `FlatList` should not be rebuilt because of the PagerView lifecycle.

## Actual behavior

With PagerView 9.x on Android, returning to the Top Tab screen causes the page/scene to be recreated, resulting in the `FlatList` being rebuilt.

## Additional investigation

We verified that the behavior is not caused by:

* `useFocusEffect`
* API refetching
* application cache
* explicit refresh logic

A/B testing consistently shows:

```text
8.0.2 → works correctly
9.x   → issue occurs
```

Since PagerView 9 introduced changes to the Android implementation, this may be related to the new Android lifecycle/page handling.

## Environment

```text
React Native:
@react-navigation/native:
@react-navigation/material-top-tabs:
react-native-tab-view:
react-native-pager-view: 9.x

Android version:
Device/Emulator:
New Architecture: Yes/No
```

## Workaround

Pinning PagerView to `8.0.2` resolves the issue:

```json
"react-native-pager-view": "8.0.2"
```

GaelCO:
## Summary
- react-native-screens fully removes and re-adds a covered screen's Fragment instead of just hiding it, destroying that Fragment's view-tree Lifecycle.
- Compose's default disposal strategy keys off that ambient Lifecycle, so the pager's `ComposeView` was torn down and rebuilt on every cover/reveal cycle: this caused a blank pager after covering the screen (#1103) and reset each page's native view state, e.g. FlatList scroll position (#1104).
- Giving the `ComposeView` its own self-owned `Lifecycle` (only destroyed for real in `onDropViewInstance`) lets the composition and each page's host survive the cover/reveal cycle untouched.

Fixes #1103
Fixes #1104

## Test plan
- [x] Manually verified with a `react-navigation` material-top-tabs stack that covering/revealing a screen containing the pager no longer blanks it out.
- [x] Manually verified a FlatList page inside the pager keeps its scroll position after the screen is covered and revealed again.

### DIFF
--- a/.maestro/tests/material_top_bar_example.yaml
+++ b/.maestro/tests/material_top_bar_example.yaml
@@ -18,6 +18,34 @@ appId: com.pagerviewexample
 - assertVisible:
     text: 'Tab1'
 
+- tapOn:
+    id: 'material-top-bar-scroll-list-button'
+
+- extendedWaitUntil:
+    visible:
+      id: 'material-top-bar-list-item-30'
+    timeout: 5000
+
+- tapOn:
+    id: 'material-top-bar-open-detail-button'
+
+- extendedWaitUntil:
+    visible:
+      id: 'material-top-bar-detail-screen'
+    timeout: 5000
+
+- pressKey: Back
+
+- extendedWaitUntil:
+    visible:
+      id: 'material-top-bar-tab-1'
+    timeout: 5000
+
+# The existing ComposeView and AndroidView host must survive the native-stack
+# cover/reveal cycle, including the FlatList's native scroll position.
+- assertVisible:
+    id: 'material-top-bar-list-item-30'
+
 - tapOn: 'Tab2'
 
 - extendedWaitUntil:
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -246,6 +246,9 @@ dependencies {
   implementation "androidx.compose.foundation:foundation:$compose_version"
   implementation "androidx.compose.runtime:runtime:$compose_version"
   implementation "androidx.compose.ui:ui:$compose_version"
+  // Needed for ComposeViewLifecycleOwner (see ComposePagerView.kt), which
+  // gives the ComposeView a self-owned Lifecycle instead of the ambient one.
+  implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.6.1"
 }
 
 if (isNewArchitectureEnabled()) {
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -238,6 +238,7 @@ repositories {
 
 def kotlin_version = getExtOrDefault('kotlinVersion')
 def compose_version = getExtOrDefault('composeVersion')
+def lifecycle_version = getExtOrDefault('lifecycleVersion')
 
 dependencies {
     //noinspection GradleDynamicVersion
@@ -248,7 +249,7 @@ dependencies {
   implementation "androidx.compose.ui:ui:$compose_version"
   // Needed for ComposeViewLifecycleOwner (see ComposePagerView.kt), which
   // gives the ComposeView a self-owned Lifecycle instead of the ambient one.
-  implementation "androidx.lifecycle:lifecycle-runtime-ktx:2.6.1"
+  implementation "androidx.lifecycle:lifecycle-runtime-ktx:$lifecycle_version"
 }
 
 if (isNewArchitectureEnabled()) {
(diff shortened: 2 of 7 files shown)


troZee:
@GaelCO 
Since v9, Android PagerView uses Jetpack Compose. When a native-stack screen covers a screen containing PagerView, react-native-screens removes and later recreates the Fragment’s view tree. This destroys the old view-tree LifecycleOwner.

If ComposeView uses that lifecycle directly, its composition is disposed or becomes unusable, resulting in the blank pager reported in #1103. Recreating the entire ComposeView fixes the blank state, but also recreates every page’s native host, causing state such as FlatList scroll position to be lost (#1104).

The fix gives ComposeView a stable, pager-owned lifecycle. It:
  - follows the active host through CREATED, STARTED, and RESUMED;
  - moves back to CREATED when detached;
  - ignores destruction of a temporary Fragment view lifecycle;
  - reconnects to the replacement lifecycle after reattachment;
  - reaches DESTROYED only when React Native actually drops the PagerView.

This is also an improvement over the previous implementation, which treated every attached view as RESUMED. A view can remain attached while its Activity or Fragment is paused or stopped, so lifecycle-aware content could otherwise keep running in the background.

I ran the new ComposeViewLifecycleOwnerTest suite successfully.

Before 
https://github.com/user-attachments/assets/c112820a-523c-412d-8d23-6efb95cb628a

After
https://github.com/user-attachments/assets/ce8e478e-0a98-4386-9bdd-965969187374

PR #1145, opened 2026-09-13
fix(android): preserve pager content during screen exit transitions

roitium:
# Summary

On Android, navigating back from a native-stack screen containing PagerView can make the pager content disappear before the screen's exit animation finishes. The screen background and views outside the pager remain visible while the pager becomes blank.

React Native removes the pager's children and calls `onDropViewInstance` while `react-native-screens` is still retaining the native view hierarchy for the exit transition. Updating the Compose page list and destroying the composition at that point clears the content too early.

This change:
- Retains the displayed page views during `startViewTransition`, while keeping React Native's logical child list up to date.
- Defers composition disposal until `endViewTransition`, with window detachment as a cleanup fallback.
- Preserves immediate page removal outside a transition.
- Stops dispatching pager events after the React Native view has been dropped.

This affects only the Android Compose implementation.

## Test Plan

- The fix was manually verified in my Android app.

Before this fix: https://github.com/user-attachments/assets/3ab80ccc-efda-4c53-a2ee-511e689dd4fe
After this fix: https://github.com/user-attachments/assets/4a8aaa55-cf4d-407f-8621-d6d7c9ab47e9

- Android with `react-native-pager-view` after v9.0.3

1. Navigate to a screen containing a PagerView.
2. Navigate back and observe the screen's exit animation.

## Compatibility

| OS      | Implemented |
| ------- | :---------: |
| iOS     | unchanged |
| Android | ✅ |

## Checklist


roitium:
This fix introduced a new crash: `java.lang.IllegalStateException: Cannot remove child at index 8 from parent ViewGroup [5178], only 9 children in parent. Warning: childCount may be incorrect!`

I will try to fix it again.

roitium:
fixed.

